PR #160, opened 2026-07-10
feat(agent-chat): detect silently-dead runs — child-exit terminal event, stall watchdog, interrupted state + Continue affordance

### LINKED ISSUE #154: feat(agent-chat): detect silently-dead runs — child-exit terminal event, stall watchdog, interrupted state + Continue affordance
## Problem

When an agent run dies without emitting a terminal event — sidecar killed by laptop sleep/lid-close, crash, or a provider usage-limit cutoff — every "working" indicator stays live forever:

- `pane_statuses` stays `Working` (sidebar spinner keeps spinning),
- the thread `streaming` flag stays true,
- the background-browser chip stays `LIVE · agent is navigating…` — its release depends entirely on the run settling to `Review`/`Idle` via a live provider event (`publish_pane_status` → `release_detached_agent_browser`, `src-tauri/src/commands/agent_chat.rs:2137-2162`; PR #139 explicitly scoped daemon-death detection out).

There is **no heartbeat, `last_activity`, stall, or timeout logic anywhere in agent chat** (grep-confirmed, Rust + TS). The user has to *know* the run is dead and type "continue" — the UI gives no indication the run stopped.

## Fix (Phase 2 of `docs/plans/run-state-truth.md`)

1. **Child-exit → synthetic terminal event.** When a provider child/sidecar exits or its stdout closes without a terminal event, emit `SessionStateChanged::Closed`/`Error` through `forward_event` so the tracker, sidebar dot, browser release, and frontend reducer all settle through the existing pipeline (t3code's `session.exited` pattern).
2. **`last_event_at` + stall watchdog.** Stamp per-thread last-event time in `forward_event`; a periodic sweep flags a mid-turn thread silent past a generous threshold (~10 min default — long quiet tool calls exist) as **stalled**: a new additive `ProviderRuntimeEvent` the UI renders as amber "no activity for Xm — the agent may have stopped", *without* killing the session. A monotonic-vs-wall-clock jump (wake from sleep) triggers an immediate pass. Optionally probe the child PID to upgrade "may have stopped" → "has stopped".
3. **Interrupted turn + Continue affordance.** A thread whose last persisted turn never settled (no `turn_completed`, no live session) renders a "Run interrupted" divider at the transcript tail plus a one-click **Continue** chip on the composer, replacing the manual "continue" users type today.

## Prior art

t3code (pingdotgg/t3code) has a first-class `interrupted` status + automatic turn settlement whenever a session leaves `running`, a `lastSeenAt` + 5-min reaper (30-min idle threshold), and wake/refocus snapshot resync — but its reaper deliberately skips mid-turn sessions and it has no stall UI, no rate-limit surfacing, and no explicit resume affordance. This issue covers exactly the cases it leaves open.

Depends on / pairs with #153 (frontend settlement + hydrate reconciliation). Design details and open questions: `docs/plans/run-state-truth.md`.

Zeus-Deus:
Closes #154.

## Problem

When an agent run dies without emitting a terminal event — sidecar killed by laptop sleep/lid-close, a crash, or a provider usage-limit cutoff — every "working" indicator stays live forever: `pane_statuses` stays `Working`, the thread `streaming` flag stays true, and the background-browser `LIVE` chip never releases. There was no heartbeat/stall logic anywhere, and the user had to *know* the run was dead and manually type "continue".

### 1. Synthetic child-exit terminal event
- The Claude and Codex per-session child-exit watchdogs now recover the in-flight `active_turn` before clearing it and emit a synthetic `TurnCompleted { status: Error { subtype: "child_exited" } }` **before** the existing `SessionStateChanged::Error`, via a shared pure helper `child_exit_events` (`agent_provider/events.rs`). Ordering is load-bearing: the turn settles through the whole existing pipeline (tracker, sidebar dot, browser release, frontend reducer) while thread state is intact, and — because `turn_completed` persists — hydrate after a restart replays a **settled** turn instead of a dangling one.
- OpenCode previously had **zero** death detection: the SSE listener gave up after 5 reconnects with only a `RuntimeWarning`. It now runs the same terminal sequence and flips a shared `dead` flag so the provider's `has_session` reports the session absent — the next send transparently rebuilds it through `ensure_live_session` (resume cursor included) instead of POSTing to a dead server.
- A child that dies while **idle** emits only the `Error` state — no spurious interrupted signal.

### 2. Stall watchdog (`RunStalled`)
A child that is alive but silent (usage-limit cutoff, wedged tool call) emits nothing, so a heartbeat is needed:
- New Tauri-managed `RunActivityTracker` (mirrors `SubagentTracker`) stamps a wall-clock last-event per thread inside `forward_event` via the pure `activity_update`: mid-turn events stamp, approval prompts **pause** the clock (waiting on the user is expected silence), settlement removes the entry.
- `spawn_stall_watchdog` sweeps every 30s and emits a new additive `ProviderRuntimeEvent::RunStalled { thread_id, silent_for_secs }` for any mid-turn thread silent ≥ 600s. Wall clock on purpose: a suspend/resume window counts as silence, which is exactly the wake-from-sleep case.
- Advisory only: never persisted, sidebar dot untouched (`map_event_to_pane_status` → `None`), re-emitted per tick to keep the duration fresh, cleared by the reducer on the next real activity. Renders as an amber "No activity for Nm — the agent may have stopped." tail notice (design tokens only).

### 3. Interrupted turn + Continue affordance
- New `interrupted` flag on `ChatThreadState`: computed at hydrate by the pure `lastTurnUnsettled` (last persisted `user_message` envelope with no later `turn_completed`) OR a replayed `child_exited` completion; set live on a `child_exited` turn error (and belt-and-braces on `session_state_changed{error}` while streaming). A user-initiated Stop produces subtype `interrupted`, so it never triggers the affordance.
- Renders a "Run interrupted" tail divider (clones the SessionStartMarker hairline pattern) and a one-click amber **Continue run** chip in the composer strip that sends the fixed text "Continue" through the normal send path — identical optimistic-bubble/rollback behavior, backend auto-resume included.
- Clears on the next running/dispatched/content event or user send.

## Docs
- `docs/features/agent-chat.md` § "Dead-run detection & interrupted turns"
- `docs/core/STATUS.md` entry
- Dev mock exposes `__codemuxChatMock.streamRunStalled()` / `.interruptRun()` for browser QA.

## Verification
- `cargo test` — full suite green: 1985 lib + all integration binaries (incl. strengthened Claude/Codex watchdog ordering tests asserting the synthetic `child_exited` completion precedes the `Error`, new OpenCode SSE give-up tests, pure-fn tracker/sweep tests, serde round-trip).
- `npm run check` — clean.
- `npm run test` — 2412 passed / 164 files (new reducer, hydrate, MessageList, and Composer cases for the notice, divider, and chip).
- `npm run verify` pipeline components all green.


Zeus-Deus:
The event plumbing, persistence policy, hydrate logic, and stall heuristics all check out, but the recovery affordance itself only works after an app restart. The live (no-restart) "Continue run" path is broken for all three providers because dead sessions are never evicted from the provider session maps. Two blocking issues:

**1. OpenCode: auto-rebuild dead-ends on `start_session`'s `contains_key` guard — every send on the thread fails permanently.**

`has_session` (src-tauri/src/agent_provider/opencode/agent.rs:245-255) reports a dead session as absent, and `emit_give_up_terminal` (src-tauri/src/agent_provider/opencode/sse.rs:462-467) sets the `dead` flag — but nothing removes the dead `OpenCodeSession` from the `sessions` map (the only removal is `stop_session`, agent.rs:233-243). On the next send, `ensure_live_session` (src-tauri/src/commands/agent_chat.rs:895-1002) sees `has_session() == false` and calls `start_session`, which checks `sessions.contains_key(&thread_id)` (agent.rs:138-148) — the corpse is still there — and returns `ValidationError("opencode session already exists for thread …")`. The resume-retry hits the same guard, so `agent_chat_send_turn` fails before the turn is attempted.

Repro: OpenCode server dies mid-turn → SSE gives up → user clicks "Continue run" (or types anything) → toast "Failed to send turn: opencode session already exists for thread …", forever, until the user stops the session or restarts the app. This is the exact flow the PR description says now "transparently rebuilds … instead of POSTing to a corpse".

Fix: evict the dead session from the map when the give-up fires (or have `start_session` remove-and-replace an entry whose `is_dead()` is true). Also note that even a successful rebuild drops the cursor: `OpenCodeSession::start` (agent.rs:149-156) never receives `input.resume_cursor`, so the "resume cursor included" claim doesn't hold as written.

**2. Claude/Codex: the Continue chip sends into the dead sidecar still occupying the session map, fails, and destroys its own affordance.**

After the child-exit watchdog fires (src-tauri/src/agent_provider/claude/session.rs:1155-1200, codex/session.rs:876-915), the session remains in the provider map with status `Error` — `has_session` is a plain `contains_key` (claude/mod.rs:316-317, codex/mod.rs:315-316) and only `stop_session` removes entries. While the app stays open (the flagship laptop-sleep/crash scenario), clicking "Continue run" (src/components/chat/AgentChatPane.tsx:1195-1197) → `ensure_live_session` is a no-op → `send_turn` → `do_send` → `JsonRpcChild::request` fails fast on the dead child (src-tauri/src/json_rpc_child/mod.rs:429-432). The user gets "Failed to send turn: send-turn RPC failed …" — and the optimistic append already cleared `interrupted` (src/lib/agent-chat/reducer.ts:845-852) while the rollback path (AgentChatPane.tsx:1163-1172) never restores it, so the "Run interrupted" divider and chip vanish after the failed click, leaving no recovery affordance and no working send path short of Stop or an app restart.

Repro: Claude sidecar killed mid-turn → "Run interrupted" + chip appear (correct) → one click → error toast, chip gone, thread still unusable.

Fix: on unintentional child exit, remove the session from the provider map (mirroring the OpenCode fix) so `ensure_live_session` rebuilds with the resume cursor exactly as it does after a restart; and/or restore `interrupted` on send-failure rollback.

**Worth addressing while in here (non-blocking):**

- A success `turn_completed` never clears `interrupted` (src/lib/agent-chat/reducer.ts:1131 spreads state). Combined with the stdout-reader/exit-watchdog race (a sidecar that emits its final result then dies can produce both a synthetic `child_exited` and a real success completion for the same turn), a successfully-finished run can be labeled "Run interrupted" — persistently, since both events persist. Clearing `interrupted` on success completion makes the race harmless.
- The Continue chip runs the full normal send path: it wipes the user's in-progress composer draft (`appendUserMessage` resets `inputDraft`, src/stores/agent-chat-store.ts:327-328), drops staged attachment chips, and actually sends staged images alongside the literal text "Continue" (`buildImagePayloads(liveAttachments)` isn't filtered, AgentChatPane.tsx:1124).
- `handleStop` fires the interrupt without awaiting and immediately tears the session down (AgentChatPane.tsx:1642-1652), so the `turn_completed{subtype:"interrupted"}` persist can lose the race and a deliberate Stop can hydrate as "Run interrupted" after restart.
- OpenCode arms `turn_active` before the POST in `send_turn`; a failed POST leaves it armed, so a later SSE give-up synthesizes a `child_exited` completion for a turn that never started.

Zeus-Deus:
Re-verified e20a2db end-to-end. The Claude/Codex live recovery is genuinely fixed now: the watchdog-set `dead` flag is correctly scoped (not keyed off `SessionStatus::Error`, never set by user Stop), `start_session` evicts-and-replaces atomically under the write lock with no false-eviction path for live sessions, the evicted corpse's `shutdown()` is hang-safe on a dead child, and the new integration tests pin the death → interrupted UI → Continue → rebuild-with-resume-cursor → send flow. The frontend fixes also all check out (rollback restores `interrupted` including the idx<0 path, success completion clears it which settles the double-completion race without breaking any legit state, Continue preserves the draft/chips and sends `images: []`, awaited Stop is bounded by the RPC timeout). All four earlier non-blocking items are resolved, and the merge with main threads correctly through the #158 settlement layer and #159 reducer changes.

One blocking issue remains, on the OpenCode path only:

**OpenCode: the rebuilt session targets a dead server — `ensure_running` never respawns a crashed `opencode serve`, so live recovery still fails in the primary scenario.**

`OpenCodeServerManager::ensure_running` (src-tauri/src/agent_provider/opencode/manager.rs:82-95) short-circuits on the cached `Some(Arc<OpenCodeServer>)` with no liveness check, and nothing in production ever calls `manager.stop()` (only the manager tests do). The SSE listener only exhausts its ~23s reconnect budget when the server is unreachable — which for a localhost child spawned by this app effectively means the `opencode serve` process died. In that case the new eviction/rebuild path runs but: `OpenCodeSession::start` → `manager.ensure_running()` returns the stale handle to the dead server → the `session_is_addressable` probe fails (falls back to fresh) → `POST /session` fails with a connect error (opencode/session.rs:113-141) → `start_session` errors → `ensure_live_session`'s fresh retry hits the identical path. Every Continue click (and every typed message) fails with "Failed to send turn: session_create_send_failed: …" until the app restarts.

The readopt path only succeeds in the narrow case where the server process is alive but the SSE stream alone broke — which is not the laptop-sleep/crash case this PR targets, and it contradicts the claim that eviction "readopts the persisted server-side session id … with fresh-session fallback". (The interrupted flag now survives the failed clicks thanks to the rollback restore, so the affordance persists — it just never succeeds.)

Fix: make `ensure_running` verify the cached child is alive (e.g. `try_wait` on the child, or a cheap HTTP probe against the cached handle) and respawn when it isn't. Since OpenCode persists sessions on disk, the existing `session_is_addressable` readopt would then recover the conversation context on the respawned server, completing the intended flow.

PR #171, opened 2026-07-15
fix(browser): default the desktop-size background-browser peek viewport to on

Zeus-Deus:
## Problem

The desktop-size viewport for the agent-chat background-browser peek (PR #140, shipped v0.13.0) is opt-in and **off by default**. Out of the box the peek popover still syncs the browser's CDP viewport to its tiny pixel size, so pages render zoomed-in / phone-width and agents have to re-send `viewport` every turn to work at desktop size.

## Change

Flip `agent_chat.background_browser_desktop_viewport` to **default ON**:

- `settings_sync.rs`: `#[serde(default = "default_true")]` on the field + manual `Default` impl for `AgentChatSettings` (checkpoints stay off by default)
- Frontend mirrors: `DEFAULT_SETTINGS`, `selectBackgroundBrowserDesktopViewport` fallback, settings-view switch fallback, dev mock
- Tests updated: legacy blobs missing the field now deserialize to ON; a new assertion covers that an explicitly saved `false` still wins; round-trip fixtures use `false` as the non-default value
- Docs: `docs/features/browser.md` § Desktop-size peek viewport and `docs/features/settings-sync.md` defaults table

## Behavior notes

- Settings blobs that **explicitly** stored `false` (any blob re-saved by a v0.13.x client) keep the container-sync behavior — deliberate opt-outs are preserved. Flipping the toggle once in Settings → Agent applies the pin.
- Blobs missing the field (pre-v0.13.0) and fresh installs get the desktop-size peek immediately.

## Verification

- `cargo test` settings_sync: 19/19 pass
- `tsc --noEmit` clean; `vitest`: 2777/2777 pass
- Visual: dev UI (Tauri mock) shows Settings → Agent → "Desktop-size background browser" ON with no stored setting

### DIFF
--- a/docs/features/browser.md
+++ b/docs/features/browser.md
@@ -155,12 +155,13 @@ own compact header instead. Both props are additive and default to today's
 pane behavior (`workspaceId` unset falls through to the exact prior
 `browser_id`-only lookup; `hideToolbar` defaults to showing the toolbar).
 
-**Desktop-size peek viewport (opt-in)**: by default, `BrowserPane` syncs the
-browser's real CDP viewport to its container — inside the 440×300 peek that
-means pages reflow at popover dimensions and agents end up re-sending
-`viewport` every turn. Settings → Agent → "Desktop-size background browser"
-(`UserSettings.agent_chat.background_browser_desktop_viewport`, synced
-settings, default OFF) instead pins the peek's viewport to 1280×800 —
+**Desktop-size peek viewport (default ON)**: with the setting off,
+`BrowserPane` syncs the browser's real CDP viewport to its container — inside
+the 440×300 peek that means pages reflow at popover dimensions and agents end
+up re-sending `viewport` every turn. Settings → Agent → "Desktop-size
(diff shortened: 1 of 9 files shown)


Zeus-Deus:
Merged — implementation verified correct across both layers (Rust serde default + manual Default impl, all four TS fallback sites, dev mock, tests all agree on `true`; explicit `false` still wins).

One follow-up for a doc agent:

**`docs/core/STATUS.md` line ~33 is now stale.** It still describes this feature in present tense as "an opt-in desktop-size viewport" with "default OFF, legacy blobs default off" — all three claims are false after this PR (`default_true`, default ON, legacy blobs missing the field default on). The canonical feature docs (`docs/features/browser.md`, `docs/features/settings-sync.md`) were updated, but STATUS.md was missed.

Known limitation (documented in the PR description, no action needed): v0.13.x users whose settings blob got a full-struct write have a baked-in `false` and stay opted out — indistinguishable from a deliberate opt-out with a plain bool.

PR #172, opened 2026-07-15
Agent chat: provider-native slash commands + /model and /default built-ins

Zeus-Deus:
## Summary

The agent chat composer's typed-`/` popup previously only offered Codemux-local commands (`/plan`, `/ask`, `/debug`, `/workflow`, `/<skill>`). None of the provider CLI's default commands were reachable from the GUI. This PR closes that gap the data-driven way — commands are **discovered live from the provider, never hardcoded**.

### COMMANDS group — provider-native slash commands
- New sidecar `list-commands` JSON-RPC method: a transient SDK `query()` probe (same lifecycle as `list-models`, but with `settingSources: ["user","project","local"]`) that returns `supportedCommands()` — the deployed CLI's built-ins (`/compact`, `/clear`, `/init`, `/review`, `/context`, …) **plus** custom `~/.claude/commands` and `<cwd>/.claude/commands` entries.
- Rust: `agent_provider/claude/slash_commands.rs` — per-cwd app-lifetime cache (errors not cached, so transient failures retry), case-insensitive dedupe, exposed via the new `list_chat_slash_commands` Tauri command. Codex/OpenCode resolve to an empty list — they expose no command-discovery surface, so the composer shows only Codemux's own built-ins there (parity with reference multi-provider clients).
- Frontend: `provider-commands-store.ts` (lazy load on first popup open, 60s TTL, keyed `(provider, cwd)`), `buildProviderCommands` with a reserved-name filter — collisions with modes / `/workflow` / `/model` / skill names are dropped so the local behaviour wins.
- Selecting a provider command inserts the literal `/name ` text (same mechanics as skills/`/workflow`); the text is forwarded verbatim and the provider interprets the leading slash itself. Codemux never executes provider commands locally. Argument hints ride along as the fallback description.

### New GUI-local built-ins
- **`/model`** (SETTINGS group): strips the typed text and pops the footer's model picker via an incrementing `openSignal` prop on both `ModelPicker` and `MultiProviderModelPicker`; the composer skips its usual textarea refocus for this pick so the popover isn't dismissed by focus-outside.
- **`/default`** (MODES group): returns to normal build mode; listed only while a non-default mode pill is active.

All pre-existing commands (modes, `/workflow`, skills) are unchanged. The dev mock serves a representative command set so `npm run dev` renders the group.

## Testing

- `cargo check` + `cargo test` (new dedupe/decode unit tests in `slash_commands.rs`)
- Sidecar `bun test` — 75/75, `tsc` (only the pre-existing SDK-type-drift error in `test/permissions.test.ts`)
- `npm run check` + `npm run test` (new builder tests; composer arrow-key wrap + footer tests updated for the new rows; 1334 chat-related tests green)
- Live end-to-end sidecar probe against the deployed CLI returned **51 commands** (all built-ins + custom entries)
- Browser-pane visual QA: popup group rendering, `/comp` → `compact` filtering, literal-text insertion, `/model` picker open, `/default` mode drop

## Docs

- `docs/features/agent-chat.md` — slash popup section rewritten with the five groups + discovery architecture
- `docs/core/STATUS.md` — branch entry added


Zeus-Deus:
Reviewed — not merging yet. The feature is solid end-to-end (Rust/TS wire shapes, serde casing, command registration, dedupe/reserved-name filter, popup integration, and the new tests all check out), but there's one real bug to fix first:

## Bug: model picker can pop open spontaneously

**`src/components/chat/pickers/ModelPicker.tsx` ~L82 and `src/components/chat/pickers/MultiProviderModelPicker.tsx` ~L198** — both have:

```ts
useEffect(() => {
  if (openSignal && !disabled) setOpen(true);
}, [openSignal, disabled]);
```

Because `disabled` is in the deps and there's no tracking of the last-consumed signal, once `openSignal > 0` (i.e. `/model` has been used once in the pane's lifetime), **every** `disabled` transition back to `false` re-runs the effect and reopens the picker. `disabled` comes from `!sessionReady` (`threadId != null && !starting && !restarting`, AgentChatPane.tsx ~2313), which genuinely toggles on session start/restart.

**Repro:** use `/model`, close the picker, then trigger a session restart (e.g. provider change) — when `restarting` flips back to false, the picker opens by itself. Related variant: using `/model` while controls are disabled silently defers the open until the session becomes ready, long after the user's intent.

**Fix:** consume the signal once — track the last-seen signal in a ref and only open on an actual increment:

```ts
const lastSignal = useRef(openSignal);
useEffect(() => {
  if (openSignal !== lastSignal.current) {
    lastSignal.current = openSignal;
    if (!disabled) setOpen(true);
  }
}, [openSignal, disabled]);
```

Apply to both picker variants.

## Nits (optional, don't block)

1. **`src-tauri/src/agent_provider/claude/slash_commands.rs`** — `invalidate()` is `#[allow(dead_code)]` and never called, so the frontend store's 60s TTL is cosmetic: after the first harvest, `.claude/commands` entries added mid-run never appear until app restart. Either wire up invalidation or drop the TTL so the code doesn't imply freshness that doesn't exist.
2. **`src/components/chat/Composer.tsx` ~L1561 / `findTriggerAtCursor`** — the COMMANDS group offers provider commands for a `/` typed mid-message, but providers only interpret leading slash commands, so mid-message insertion degrades to literal prose. Consider a leading-position guard for the provider-command group.
3. **`src/components/chat/Composer.test.tsx`** — unlike `Composer.skills.test.tsx`, it doesn't mock `listChatSlashCommands` or reset the provider-commands store, so popup tests exercise the new rows via an unmocked IPC failure path (error entry persists across tests in the file). Add the mock + `invalidate()` reset for deliberate coverage.

Once the `openSignal` fix lands, this is good to merge.

PR #174, opened 2026-07-16
perf(chat): stage images at attach time and make the first send feel instant

Zeus-Deus:
## Problem

1. **Sending a prompt with several pasted screenshots took minutes.** Image bytes crossed Tauri IPC as a JSON `number[]` — roughly 4 bytes of JSON per image byte, so five ~2-3 MB screenshots became a ~40-60 MB JSON string built in JS, marshalled through WebKit, and serde-parsed in Rust, all inside `agent_chat_send_turn`.
2. **The first send in a new thread always sat on a dead composer for seconds.** The draft surface only flipped after workspace/worktree creation (including 5-9s AI branch naming), a 5s cwd poll, pane creation, session start (up-to-8s MCP prime + sidecar spawn), and the send itself — all serial, all invisible.
3. **Image thumbnails in the sent thread rendered as broken-image icons after hydrate.** Persisted images live under `~/.config/codemux/agent-chat/images/`, and the asset-protocol scope `["**"]` never matched them — Unix glob scopes skip leading-dot path components by default, so every `convertFileSrc` load 404'd.

### Attach-time image staging (kills the multi-minute send)
- New raw-body `agent_chat_stage_image` command (bytes as the invoke's raw request payload, MIME on an `x-media-type` header — the fs-plugin `write_file` mechanism) writes each image to a staging dir the moment the chip's bytes resolve, while the user is still typing. A base64-JSON body fallback covers the web-remote transport, which cannot carry raw invoke bodies (`isRemoteClient()` skips the doomed raw attempt entirely).
- `agent_chat_send_turn` now takes `images: [{ path, media_type }]` refs; `finalize_chat_images` renames staged files into the thread's dir (path-classification security gate: staging or own-thread dir only) and reads bytes for the provider — the adapter `SendTurnInput`/`ImageInput` contract is unchanged.
- Removed chips fire `agent_chat_discard_staged_image` (staging-contained delete); a 24h-grace startup sweep reaps leaked staging files; all image fs ops are `tokio::fs` (the old blocking `std::fs::write` inside the awaited send path is gone).

### Instant first-send feedback
- `materializeAndSend` appends the client-nonce'd optimistic user bubble at the top — before any workspace/session work — and rolls it back via `removeUserMessageByNonce` on every failure branch. `DraftChatSurface` immediately swaps its landing for a pending conversation: the sent bubble (with `data:`-URL thumbnails) plus a shimmer status line advancing through the real phases via the new `onPhase` hook; the composer clears at once and is restored on failure.
- The pane's deferred-worktree first-send path mirrors the warm-send pattern: optimistic bubble + composer clear first, nonce rollback on failure.
- `create_empty_workspace` / `create_worktree_workspace` return `WorkspaceCreated { workspace_id, cwd }`, so the 200ms×5s `waitForWorkspaceCwd` poll is now fallback-only (legacy string-shape tolerated; existing callers untouched via wrapper delegation).
- Fire-and-forget `agent_chat_prime_mcp` on draft/new-thread mount overlaps MCP warmup with typing; a registry-level `prime_lock` serializes it against `agent_chat_start_session`'s backstop prime so MCP children can't double-spawn.

### Thumbnail rendering
- `assetProtocol.scope` → `{ "allow": ["**"], "requireLiteralLeadingDot": false }` so hydrated fs paths under dot-directories actually serve.
- Belt-and-braces `useImageWithFallback` in `UserMessage`: when an fs-path `` errors (dev mock, web-remote — no asset protocol), read the bytes back over IPC (`agent_chat_read_image`, canonicalized-path containment inside the chat-image root) and render a cached blob URL before ever showing the placeholder.

## Verification

- `cargo check` + `cargo test` (2184 passed; 12 new unit tests covering MIME validation, extension mapping, path containment/traversal rejection, staged-vs-thread classification, and the finalize rename; the only 2 failures in a fully-loaded parallel run were the live-`opencode`-binary environment tests, which pass in isolation and are untouched by this change)
- `npm run check` + `npm run test` (2790/2790, incl. new `image-staging`, `UserMessage` fallback-loader, and reordered-materialize suites)
- Dev-mock browser drive: paste → chip stages via the raw-body command → send → optimistic bubble renders its thumbnail, zero console errors; warm-path send with an image verified the same way.

Docs updated per handoff discipline: `docs/features/agent-chat.md` (attachments/staging, first-send ordering, cwd invariant, constraints) and `docs/core/STATUS.md`.


Zeus-Deus:
Merged — the hard parts all verified correct: the send-before-staging race (`awaitImageStaging` + fresh store re-read + unstaged gate with rollback + toast), nonce rollback symmetry on every failure branch, the queued-turn image bridge, and the Rust path-containment boundary (canonicalized parent, `..`/sibling/cross-thread refs rejected, unit-tested). Conflicts with #172 (two doc files) were resolved on the branch and CI re-ran green on the combined result.

Follow-ups for a cleanup agent (none were merge-blocking):

1. **Staged-file leak on mid-flight chip removal** — `src/components/chat/AgentChatPane.tsx` ~L2802 + `src/lib/agent-chat/image-staging.ts`: removing a chip while its upload is in flight never discards the staged file (`updateStagedAttachment` no-ops on the removed id, and `beginImageStaging`'s success path has no "chip still exists?" check). Same for abandoned drafts/closed panes. Only reclaimed by `sweep_stale_staged_images`, which runs at startup only with a 24h grace — long-running sessions accumulate leaks. Fix: discard-on-orphan in the staging resolve, and/or run the sweep periodically.
2. **Empty-string cwd defeats the poll fallback** — `src/tauri/commands.ts` ~L295 (`normalizeWorkspaceCreate`): backend `unwrap_or_default()` can return `cwd: ""`, which `obj.cwd ?? null` keeps, so `created.cwd ?? waitForWorkspaceCwd(...)` short-circuits on `""` and the send fails immediately instead of engaging the 5s poll. One-line fix: map `""` → `null`.
3. **Blob-URL fallback vs documented hosted CSP** — `src/components/chat/UserMessage.tsx`: the IPC-read fallback mints a `blob:` object URL, but `docs/plans/app-codemux-org-hosting.md` L157 specifies `img-src 'self' data: https:` (no `blob:`) — on the hosted web client the fallback fetches bytes then gets blocked anyway. Either add `blob:` to the documented img-src or emit a `data:` URL.
4. **Silent image drop on finalize fs errors** — `src-tauri/src/commands/agent_chat.rs` `finalize_chat_images`: rename/read failure now skips the image (`continue`) while the optimistic bubble still shows the thumbnail; previously the provider always got the in-memory bytes. Also a mid-loop security rejection leaves earlier already-renamed refs as orphans. Consider failing the turn (or surfacing a toast) instead of silently sending without the image.
5. Nits: `DraftChatSurface.tsx` submit IIFE dropped the old `.finally(() => { sendInFlightRef.current = false })` (currently unreachable wedge, but the sibling handler kept try/finally); `agent_chat_read_image` returns `number[]` JSON bytes on the fallback path (a `tauri::ipc::Response` would be symmetric with the upload); `agent_chat_prime_mcp` primes the *active* workspace's project, which can serialize a fast cross-project first send behind a wrong-project warmup via `prime_lock`.

PR #175, opened 2026-07-16
Workspace archive: one-click restorable archive replaces sidebar close

Zeus-Deus:
## What

Replaces the sidebar's X/Hide/Close flow with a two-tier model: **archive** (frictionless, always reversible, never touches disk) as the primary action, and **delete** demoted behind an explicit destructive confirm. Adds a **Settings → Archive** panel to restore anything back.

## Why

Closing a workspace was irreversible from the UI — "Hide" recorded nothing, so recovery meant manually recreating the worktree. And the root-workspace protection only existed as a hidden hover button: the context menu and MCP path could still close the root, and the delete path hardcoded `force=true` past the backend's dirty-worktree guard. This PR makes the safe action the easy one and puts the safety invariants in the backend where every surface shares them.

## Highlights

**Archive**
- One-click archive on every local row — including the protected repo root — with an Undo toast. Archiving the root leaves the project group intact (grouping keys off `project_root`); restore is one click in Settings → Archive.
- Entries persist in `AppStateSnapshot.archived_workspaces` (additive serde; old `layout.json` loads unchanged).
- Unarchive reuses the normal worktree/root creation paths — on-disk worktrees are reused, entry removed only after a successful restore, custom titles survive.
- Attach-in-place and remote (`host_id`) workspaces are refused (archiving them couldn't be lossless); those rows keep a plain non-destructive Close.
- "Close Project" → "Archive Project" (sequential, worktrees first, root last; drafts cleared only on full success).

**Safety hardening**
- Protected-root deletion refused inside `close_workspace_with_worktree_impl` — one guard covers the sidebar dialog, control socket, and MCP `workspace_close`. Shared `is_undeletable_root` predicate keeps the close and archive paths aligned (truth-table tested).
- Honest force: non-forced deletes of dirty/unpushed worktrees are refused by a git pre-flight **before** any teardown or state change, so the confirm dialog stays mounted and escalates to an explicit "Force delete". All four delete surfaces (sidebar, Settings → Archive, workspaces overview, empty-workspace state) share one escalation flow (`use-force-delete.ts`); the error wording is pinned by a Rust contract test.

**Programmatic access**
- Socket: `archive_workspace` / `unarchive_workspace` / `list_archived_workspaces`; MCP: `workspace_archive` / `workspace_unarchive` / `workspace_archive_list`.

**Docs**: new `docs/features/workspace-archive.md`, updates to `STATUS.md`, `INDEX.md`, `mcp-server.md`.

## Known scope limits

- The archive list is device-local; sync-wise, archiving looks identical to the old close (soft-delete). Cross-device archive is future work (documented in the feature doc).
- Archiving runs teardown scripts (like the old hide); a failing teardown aborts that workspace's archive and surfaces in a toast.

## Verification

- `npm run verify` green end to end: `cargo check` + `cargo test` (2196 lib tests + all integration binaries, 0 failures) + `tsc` + `npm run test` (193 files / 2821 tests, 0 failures).
- New tests: archive store ops + serde back-compat, root-refusal truth table, archivability refusals, force-wording contract test, archive-section component suite, shared escalation hook suite, and the restored safety-invariant test that shift-click on a protected root can never open the delete dialog.
- Manual browser verification (dev mock): archive root with siblings open → project group survives; unarchive → restored + activated with branch intact; dirty-worktree delete → refusal message shown verbatim, "Force delete" escalation completes; root rows expose no delete affordance anywhere (hover, shift-click, context menu, archive panel).


Zeus-Deus:
Reviewed — not merging yet. The architecture is sound and the safety hardening is a real improvement (honest force semantics, backend-enforced root protection across UI/control-socket/MCP, strong serde back-compat + test coverage). But there's one regression to fix first, plus two weaker-than-documented restore paths worth addressing or re-documenting.

## Blocker: empty-workspace "Delete workspace" is broken for every non-worktree workspace

**`src/components/layout/empty-workspace-state.tsx` ~L91–105** — `handleDeleteWorkspace` unconditionally passes `removeWorktree=true`:

```ts
closeWorkspaceWithWorktree(workspaceId, true, false, force)
```

The new backend guard (`refuse_worktree_removal` → `is_undeletable_root`, `src-tauri/src/commands/workspace.rs` ~L975) refuses whenever `worktree_path.is_none()` — not just for protected roots. On main, a non-worktree workspace (primary checkout, plain folder, Home) simply skipped worktree removal and closed. Now the only action on the empty-workspace screen always fails for that class of workspace, with a toast that's also wrong for plain-folder/Home workspaces ("This workspace is the project's root checkout…"). `EmptyWorkspaceState` renders for any workspace with zero tabs, so this hits the most common workspace kind.

**Fix:** gate on `worktree_path` like `workspace-overview-row.tsx` does with `canRemoveWorktree` — archive/close for non-worktree workspaces, delete-with-worktree only when there's a worktree to remove.

## Minor: two advertised restore paths verifiably fail

**`src-tauri/src/commands/workspace.rs` ~L1184–1199 + `src-tauri/src/git.rs` ~L2624–2643** — the "directory gone but branch survives → recreate worktree" path doesn't work. Archiving touches nothing, so a hand-deleted worktree dir leaves a stale git worktree registration; `git_create_worktree`'s reclaim/prune block only runs when `worktree_path.exists()`, so `git worktree add` dies with `fatal: '<path>' is a missing but already registered worktree`. The unarchive pre-flight passes via the branch check, then the restore fails with the raw git fatal instead of the friendly error. Same root cause: **imported worktrees at non-conventional paths** archive fine but can never restore (restore always targets `~/.codemux/worktrees/<repo>/<branch>` → `already used by worktree at <custom path>`). Never lossy (the entry is kept), but `docs/features/workspace-archive.md` and the MCP `workspace_unarchive` description claim these paths work. Fix: run `git worktree prune` / reclaim before `worktree add` in the restore path (and target the recorded `worktree_path` for imported worktrees), or narrow the docs + pre-flight.

## Nits (optional)

1. **`src-tauri/src/commands/workspace.rs` ~L805/821/840/934** — residual TOCTOU: teardown scripts run after the dirty pre-flight and execute inside the worktree; if one creates files, the inner `git_remove_worktree(force=false)` guard refuses after `state.close_workspace` already tore the workspace down — orphaned worktree, and the force-escalation re-issue hits "No workspace found". The comment claiming the dialog "stays mounted and can escalate" isn't guaranteed.
2. **`src-tauri/src/git.rs` ~L2751** — `git show-ref --quiet -- <name>` matches tags too; a deleted branch shadowed by a same-named tag passes the unarchive pre-flight, then `worktree add` checks out the tag detached. Use `--verify refs/heads/<name>`.
3. **`src/components/settings/settings-view.tsx` ~L2038** — copy says unarchive brings a workspace back "exactly where it was," but restore builds a fresh workspace (single layout, new id) — pane layout/tabs/sessions aren't preserved. Soften the copy.

Once the empty-workspace-state fix lands (and the restore-path claims are either fixed or re-scoped), this is good to merge.

PR #177, opened 2026-07-17
fix(agent-chat): self-heal stale Claude session resume instead of wedging the thread

Zeus-Deus:
## Problem

Agent Chat threads sometimes stop sending messages entirely: every turn ends with

> Turn ended: error_during_execution — No conversation found with session ID: …

and the only workaround is creating a new thread. This happens when the persisted `sdk_session_id` goes stale — the Claude CLI deletes old conversation JSONLs (periodic cleanup, version updates), but Codemux kept handing the dead id back as `resume` on every session rebuild. Nothing ever cleared the stale cursor, so the thread stayed wedged across restarts.

## Fix — three recovery layers

**1. Sidecar self-heal** (`sidecar/claude-agent/src/session.ts`)
- Detects the stale-resume `result` error (and the thrown-error variant), **suppresses it**, emits a new `resume-fallback {threadId, staleSessionId}` notification, rebuilds a fresh query with all resume state cleared, and **replays the pending user turn verbatim** — so the send just works.
- One recovery attempt per turn; a repeat failure surfaces normally (no loops).
- The CLI validates `--resume` eagerly, so the error can arrive right after session start with no turn in flight — recovery fires there too, with nothing to replay.

**2. Rust + inline notice**
- `resume-fallback` clears the in-memory cursor and translates to `ResumeCursorUpdated(null)` + a `resume-fallback:`-prefixed `RuntimeWarning`.
- The persist path clears the DB column on exactly JSON null (`clear_agent_chat_sdk_session_id`); the rebuilt query's fresh `sdk-session-id` re-persists moments later.
- The frontend classifier promotes the warning to a subtle inline transcript notice: *"Previous session context couldn't be restored, so this turn continues in a fresh session. Your chat history is preserved."*
- The replayed turn stays `Running` — no state clobber, no phantom turn end.

**3. Resume preflight**
- Before `ensure_live_session` / `agent_chat_start_session` hand a persisted Claude cursor to the SDK, they positively confirm `<config>/projects/*/<id>.jsonl` still exists (`$CLAUDE_CONFIG_DIR` or `~/.claude`). Only **definitive absence** drops the cursor (and best-effort clears the column); any IO doubt keeps it, so a valid resume is never lost to a flaky probe. Includes a path-traversal guard on the probed id.

## Verification

- **End-to-end against the real CLI**: drove the built sidecar over stdio with a deliberately bogus resume uuid — the exact production failure. Before the fix: the error result was forwarded and the turn died. After: `resume-fallback` fired, the error was suppressed, a fresh session spawned, and the turn completed with a successful result.
- Sidecar: 81 bun tests pass (6 new: recovery+replay, eager-error-before-any-turn, retry cap, unrelated errors untouched, thrown-error path, fresh-session id re-emission).
- Rust: `agent_provider::claude` 138, `commands::agent_chat` 69 (5 new preflight tests), `database` 91, `claude_adapter` integration 42 (new fake-sidecar resume-fallback test), `agent_chat_commands` integration 28 — all pass. (`git::tests` parallelism flakes are pre-existing and pass 196/196 single-threaded; module untouched.)
- Frontend: `tsc` clean, 2844 vitest tests pass (new runtime-notice promotion test).
- Docs: `docs/features/agent-chat.md` updated (self-heal section + notification mapping).


Zeus-Deus:
Reviewed in depth — the core mechanism is correct across all three layers (scoped error detection, one-shot replay per turn, cursor clear on `Null`, asymmetric preflight drop) and the tests exercise the real behaviors. Merging.

Non-blocking follow-ups worth a future pass:

1. **Start-session preflight clears the wrong row** (`src-tauri/src/commands/agent_chat.rs` ~556): every frontend flow that passes a `resume_cursor` to `agent_chat_start_session` mints a fresh thread id, so `clear_agent_chat_sdk_session_id(&input.thread_id.0)` clears a row with no stale id while the old thread's row keeps it — the dead history entry is never healed (no wedge, but it survives forever). Clearing by `sdk_session_id` value would fix it.
2. **Missing `!this.interrupting` guard on the result-message recovery path** (`sidecar/claude-agent/src/session.ts` ~355): the thrown-error path has it, the result path doesn't. If the user hits Stop in the ~1s window before the eager `--resume` validation error arrives, recovery replays the turn the user just stopped and the interrupt is silently swallowed. A second Stop recovers, but the asymmetry looks unintentional.
3. **`lastUserMessage` assigned after `ensureLiveQuery()` with an await between when `modelOverride` is set** (`session.ts` ~564): narrow race where recovery can replay a previously interrupted turn and then also run the new one. Moving the `lastUserMessage`/`resumeFallbackAttempted` assignments before `ensureLiveQuery()` closes it.
4. **Process-global `CLAUDE_CONFIG_DIR` in tests** (`src-tauri/tests/agent_chat_commands.rs` ~862): `set_var` is never restored and the binary runs tests in parallel — any future resume-exercising test in this binary can flake on it.

PR #184, opened 2026-07-18
fix(agent-chat): launch New Chat/resume sessions in the mode the UI shows

Zeus-Deus:
## Problem

Setting the composer to **Full access** didn't stick for chats started via the session-history dropdown: the agent still asked for permission on every tool, and the only workaround was re-picking Full access (already checkmarked) and typing "continue". Reported as happening only in main-workspace threads.

## Root cause

`use-agent-chat-session-actions.ts` (the hook behind the dropdown's **New Chat** and **resume-a-chat** actions) was the one launch path in the app passing `permission_mode: null` to `agent_chat_start_session`:

- The SDK booted the CLI in `default` mode → Allow/Deny card on every tool.
- The fresh store slice's footer pill defaulted to `bypassPermissions` → UI showed "Full access".
- The DB row's `permission_mode` stayed NULL (`keep_or_set(None)` never writes), so nothing corrected the drift.
- The v0.14.1 sidecar bypass-downgrade guard (#178) never fired: it keys on intended mode `=== "bypassPermissions"`, but the sidecar's intended mode read `undefined`. That fix was aimed at the wrong layer.

"Main threads only" because worktree prestart, draft materialize, and fresh-pane boot all pass a real mode — only the dropdown flows (used in long-lived main-workspace panes) passed null.

## Fix (two layers)

1. **Frontend (root cause)** — `src/hooks/use-agent-chat-session-actions.ts`
   - New Chat launches with `DEFAULT_THREAD_PERMISSION_MODE`.
   - Resume threads the record's persisted config through (`permission_mode` NULL-healed to the provider default, plus `model`/`effort`/`context_window`).
   - Both paths seed the slice's `permissionMode` **and** `sessionLaunchMode`, so the pill, restart detection, and the actual launch agree.
2. **Backend (defense-in-depth)** — `src-tauri/src/commands/agent_chat.rs`
   - `agent_chat_start_session` heals a `None` mode via the new `resolve_start_permission_mode` helper (mirrors `ensure_live_session`'s resume-path heal) before the provider launch and the row persist. No future null-passing caller can launch a session in a mode the UI doesn't display.

Docs updated (`docs/features/agent-chat.md`, `docs/core/STATUS.md`) to record the corrected diagnosis.

## Verification

- `cargo check` + `cargo test` — pass (2 new unit tests for the heal; one unrelated terminal PTY soak test flaked under full-suite load, passes in isolation and on unmodified main)
- `npm run check` — clean
- `npm run test` — 2849/2849 (5 new hook tests covering New Chat mode, resume passthrough, NULL heal, and slice seeding; 1 stale header-test assertion updated)


Zeus-Deus:
Deep-dived this before merging — the fix is verified correct for the reported bug:

- Confirmed `use-agent-chat-session-actions.ts` was the **only** production call site passing `permission_mode: null` (checked every `agentChatStartSession` caller: pane mount, draft recovery, stop-restart, `restartSessionWith`, materialize, worktree prestart — all pass a real mode).
- Frontend and backend heals agree for Claude (`DEFAULT_THREAD_PERMISSION_MODE === fallback_permission_mode(Claude) === "bypassPermissions"`), the heal runs before both the provider launch and `config_for_persist`, so the row no longer stays NULL.
- This makes the #178 sidecar bypass-downgrade guard actually effective (its intended-mode read is no longer `undefined`) — no regression there.
- Seeding both `permissionMode` and `sessionLaunchMode` correctly suppresses the spurious-restart path.
- New Rust + hook tests re-run on the branch: all pass.

Merging. Three follow-ups worth tracking (none blocking — the first is a pre-existing pattern this PR extends, not a new regression):

1. **Codex panes can still desync via the same dropdown**: the hook heals with the Claude-specific `bypassPermissions` for all providers. `codex_permission_mode_to_policy_pair("bypassPermissions")` returns `None` (Codex launches with its own prompting defaults) while the picker shows the Codex "Full access" label — and because the wire now carries `Some(...)`, the backend's per-provider `resolve_start_permission_mode` heal never fires, and `keep_or_set` persists the bogus mode so the `ensure_live_session` NULL-heal can't repair it later. A provider-aware frontend default (or passing `null` and letting the backend heal per-provider) would close it.
2. In the backend heal block, `input.permission_mode.take()` inside the `is_none()` branch always yields `None` — dead argument, worth cleaning.
3. Stale checkout at `.claude/worktrees/vast-churning-moonbeam/` contains a pre-PR copy of `AgentChatPaneHeader.test.tsx` and gets swept up by broad vitest globs → spurious local failure. Delete it.

PR #185, opened 2026-07-18
fix(chat): restore smooth transcript scrolling on WebKit engines

Zeus-Deus:
## Problem

Scrolling the agent chat transcript in the GUI felt rigid and jumpy on Linux: fast wheel scrolling seemed to decelerate, and scrolling up into history would visibly "bug back down" then correct itself.

## Root causes (both verified empirically)

**1. The MessageTrail rail's wheel forward never unpinned the scroller engine.**
`forwardWheel` wrote `viewport.scrollTop += deltaY` directly. That fires only a `scroll` event — the engine's stick-to-bottom state machine releases its pin exclusively on genuine user input (`wheel`/`touchmove`/nav keys) observed on the viewport, and its fallback direction-based release is suppressed for 180ms after every programmatic scroll (continuously re-armed during token streaming). So wheeling up with the cursor over the left rail mid-stream stayed pinned and got snapped back to the tail on every content mutation. The stale comment referenced `pinnedRef` machinery deleted when the shadcn scroller landed.

**2. The free-scroll stability design assumes native CSS scroll anchoring — which WebKit doesn't have.**
Rows use `content-visibility:auto` with `contain-intrinsic-size` estimates that settle to real heights on first reveal; the design delegates absorbing those settles to `overflow-anchor` (on exactly while `data-scrollable~=end`). Verified on WebKitGTK 2.52.5 (the Tauri Linux webview; WKWebView is the same family):
- `CSS.supports('overflow-anchor','none')` → **false** (WebKit never implemented the property)
- implicit anchoring → **absent** (a 200px block inserted above the viewport leaves `scrollTop` unchanged — content visibly jumps)
- `content-visibility` → **supported**, so the settles do happen, uncompensated

That is precisely the "drift the viewport hundreds of px per wheel batch" regression the docs describe from the earlier blanket-`overflow-anchor:none` build — permanently active on WebKit platforms.

## Fix

- **MessageTrail**: dispatch a real `WheelEvent` on the viewport (unpins the engine) before the `scrollTop` write (performs the scroll — synthetic events have no default action). Comment rewritten truthfully.
- **New `ScrollAnchoringShim`** (`scroll-anchoring-shim.tsx`, mounted in `MessageList` as a viewport sibling): reproduces native anchoring in JS — rAF-throttled capture of the topmost visible row on scroll; a ResizeObserver on the content element re-pins that row by nudging `scrollTop` when a settle moves it. Gated on the exact conditions the CSS design uses (`data-scrollable` contains `end`, no `data-autoscrolling`), so the engine still owns the tail while pinned. Feature-detected via `CSS.supports` — **strict no-op on Chromium/WebView2** — with a `codemux:anchoring-shim` localStorage override (`on`/`off`) for A/B testing, mirroring `codemux:transcript-fade`.

## Verification

- Full vitest suite: 2856 passing (12 new — trail wheel forward, shim gates/correction/overrides), `tsc` clean, `cargo check`/`cargo test` clean
- Live dev-mock (491-row transcript): forced-on shim compensated a +300px height settle above the viewport with **0px anchor drift** while unpinned; stayed inert while pinned at the tail (engine owned the snap); rail wheel now lands a real wheel event on the viewport, scrolls exactly by deltaY, and the engine stays unpinned
- Raw-`scrollTop`-write control test reproduced the original snap-back, confirming the unpin mechanism

## Not in this PR (follow-ups, by design)

- The upstream scroller's per-scroll-event O(n) `getBoundingClientRect` walk (requires vendoring/patching `@shadcn/react`; measure feel after this PR first)
- Revisiting the blanket `WEBKIT_DISABLE_COMPOSITING_MODE` default (biggest raw-smoothness lever on Linux, but it's a crash workaround — needs its own hardware-tested PR)


Zeus-Deus:
Deep-dived this against the actual scroller-engine internals before merging — both fixes verified correct:

- **Wheel forward**: the engine has no native wheel listener and no `isTrusted` check — the viewport's React `onWheel` calls `userScrollIntent` which releases the pin unconditionally, and `deltaY` is never read by the engine. Synthetic dispatch (bubbles, synchronous) reaches the React root listener before the `scrollTop` write; no double-scroll on any engine since an untrusted WheelEvent has no UA default action. The claimed failure mode is real and reproduced in engine source: the direction-based fallback is suppressed while `autoscrollingRef` is armed (180ms constant, re-armed per streaming tail snap) — so the bug was stream-specific, exactly as described.
- **Anchoring shim**: the gap is real (`reanchorToAnchoredMessage` is unreachable here since every row sets `scrollAnchor={false}`, so free-scroll stability relied entirely on CSS `overflow-anchor`, absent on WebKit). Gate polarity, sign/clamp math, epsilon (matches engine 0.5), cleanup, and mount point all check out. No feedback loops (the correction write doesn't resize content), and forced-on Chromium can't double-compensate (native anchoring corrects during layout before RO callbacks measure ~0). All engine-fight scenarios refuted: pinned-at-bottom, tail snaps, tail-append-while-reading.
- Tests re-run on the branch: 16/16, `tsc` clean. The tests are behavioral, not smoke.

Merging. Follow-ups worth tracking (non-blocking):

1. **Perf**: `captureAnchor` runs `content.querySelectorAll("[data-message-id]")` per captured frame, ungated — while pinned during streaming, every tail snap fires `scroll`, so a 500-row transcript pays an O(n) DOM query per frame for anchors the gates then discard. Cache the row list or skip capture while `"end"` is absent.
2. **Rail-tick jumps**: `scrollToMessage` enters `settling-jump` without `data-autoscrolling`, so the settle loop and the shim can both write `scrollTop` during a cold history jump on WebKit — they converge, but shim corrections can reset the settle loop's stable-frame counter (brief jitter). Consider gating the shim on `settling-jump` too.
3. Nits: the step-1 comment slightly overstates ("only releases on genuine user input" — the direction fallback does release when idle; the pin only survives inside the 180ms window); `forwardWheel` ignores `deltaMode` (pre-existing, pixel-mode everywhere shipped); the localStorage toggle caches at module level so it needs an app reload.

PR #186, opened 2026-07-19
Move right-panel toggle next to window controls in title bar

Zeus-Deus:
## What

Moves the right-panel toggle in the GUI-chrome title bar from the front of the right cluster (left of the Set Run button) to the far right, directly beside the minimize button.

## Why

Layout toggles read better docked with the window chrome than mixed in with action buttons. The right cluster now reads: Run split button / resource monitor / IDE launcher | separator | panel toggle, window controls.

## Notes

- On the web remote client the native window controls render nothing; the separator already hides itself there, so the toggle cleanly ends the row.
- UI-only change in `src/components/layout/title-bar.tsx`.

## Verification

- `npm run check` — clean
- Full frontend suite: 202 files, 2924 tests passing
- Visually confirmed in the browser pane with the dev mock


Zeus-Deus:
The layout change itself is small and visually coherent, but it leaves the canonical GUI-chrome doc stale. `docs/features/gui-chrome.md` still specifies the right cluster as `[right-panel toggle] [RunButton] [ResourceMonitor] [IdeLauncher] [sep] [WindowControls]`, while this PR intentionally moves the toggle into the window-controls group. Please update the Current Model / rehomed-controls description in that doc so `docs/` remains the source of truth.

Also note that #187 touches the same title-bar area, so this branch may need a refresh after that PR lands.

Zeus-Deus:
Implemented the requested corrections and re-reviewed the branch against current main. The conflict resolution preserves draft-workspace safety: Run, IDE, and the panel toggle are suppressed for draft chrome, while the panel toggle is placed immediately beside the native window controls for real workspaces. Added DOM-order and draft-gating regression assertions; focused title-bar/workspace tests and TypeScript checks pass, and I visually verified both workspace and draft chrome in the dev mock.

PR #188, opened 2026-07-19
Headless remote access: codemux serve + terminal remote toggle

### LINKED ISSUE #176: Headless remote access: decouple the web-remote server from the desktop GUI
## The one thing this issue delivers

> SSH into any machine — including one with no display — start CodeMux remote access from the terminal, get a pairing code, and connect to it from a browser.

That's it. Everything below is either **how** to get there (implementation phases) or **what must not break** while doing it (regression guardrails). No other feature work is in scope.

## Why it doesn't work today

1. `codemux remote pair` needs the running desktop app's control socket — if CodeMux is closed, it fails.
2. Remote access can only be enabled from the Settings GUI — there is no CLI toggle.
3. On a machine with no display, the app can't launch at all (Tauri needs a window), and — the deep problem — every remote command is replayed **through the main GUI window's webview IPC** (`dispatch.rs` → `app.get_webview_window("main")`). So a headless process couldn't serve remote commands even if it could start.

Point 3 is why this isn't just "add a launch command": making the SSH scenario work on a headless box requires decoupling the command dispatch from the GUI. Competing agent-IDE products are built server-first (headless server + CLI `serve` that prints pairing details; the desktop app is a thin client of the same server), which is why they get this scenario for free.

### Phase 1 — CLI toggle (covers the "app is running, I forgot to enable it" case)
- `codemux remote enable [--scope all|tailscale|loopback] [--port N]` and `codemux remote disable` as control-socket commands. `web_remote_pair` in `control.rs` is the template; the settings toggle path already exists app-side.
- After this: SSH into a machine where the app is running → enable → `codemux remote pair` → connect. The GUI-running case is done.

### Phase 2 — decouple invoke dispatch from the webview (prerequisite for headless)
- Replace the `get_webview_window("main")` replay in `src-tauri/src/web_remote/dispatch.rs` with a direct command registry: remote invokes call the Rust command handlers directly. Keep the ACL semantics.
- Also a robustness fix on its own: today, if the main window is gone, ALL remote transports break — including the iroh/hosted path, which shares this dispatcher.

### Phase 3 — `codemux serve` (the actual headless deliverable)
- A CLI entrypoint that runs the core (state, DB, PTY daemon, web-remote server + embedded frontend bundle, and the iroh endpoint when `relay_mode_enabled`) without creating a window. Prints pairing details on startup, honoring the same settings.
- Acceptable intermediate: `serve` and the GUI are mutually exclusive on one machine (single-instance lock already exists). Long-term nice-to-have (NOT required here): the GUI becomes a client of the same core.

## Useful groundwork already in place

- The auth layer (`web_remote/auth.rs` — pairing tokens, session hashes, WS tickets, rate limiting) and `control_pair_from` are already pure over `Shared`/`DatabaseStore`, unit-tested without an `AppHandle`. The security model needs zero changes.
- The control socket comment in `control.rs` says pairing-over-SSH was the design intent all along.

## Guardrails — existing features that share these code paths and MUST keep working

Not scope. Not new work. Just don't break them; they ride the same dispatcher/server being modified:

- **Account-mode admission** (`web_remote/account.rs`, `POST /api/pair-account`) — validates against the desktop's cached signed-in user (`load_cached_user`); a headless `serve` needs access to the same cached credential state for this to keep working.
- **Hosted web frontend** (`app.codemux.org` — `src/remote/bootstrap.tsx`, `hosted-oauth.ts`, `device-registry.ts`) — reaches the desktop exclusively via the relay transport.
- **iroh relay transport** (`web_remote/iroh.rs`) — E2E QUIC, feeds the same `handle_text_frame` dispatcher and shared `ConnectionRegistry` as `/ws`; default-off behind `relay_mode_enabled`.
- All security invariants: pairing single-use/TTL, session hashing, ticket-gated WS, origin checks, approval mode, revocation closing live sockets, account-mode user match, relay E2E encryption.

Bonus that falls out for free if the guardrails hold under `serve`: the headless box is also reachable from `app.codemux.org` (account + relay), not just LAN/Tailscale. No extra work should be done to chase this — it should simply keep working.

## Acceptance

- On a machine with no display: `ssh box`, `codemux serve` → prints pairing URL + QR → phone/browser pairs → full workspace/terminal/agent control.
- On a machine with the GUI running: `codemux remote enable` + `codemux remote pair` works without touching Settings.
- Nothing in the guardrails list regresses.

## Non-goals
- No new cloud infrastructure (existing `api.codemux.org` / `app.codemux.org` / relay are untouched, just not broken).
- No mobile app.
- No weakening of the auth model to make headless easier.

Zeus-Deus:
Closes #176.

SSH into any machine — including one with no display — start remote access from the terminal, get a pairing code, and connect from a browser.

### Phase 1 — CLI toggle (app running, GUI not needed)
- `codemux remote enable [--scope all|tailscale|loopback] [--port N]` / `codemux remote disable` as control-socket commands, reusing the exact Settings-pane code paths (config persistence, bind rollback on failure, the tailscale-no-tailnet hard error — never a silent 0.0.0.0 fallback).

### Phase 2 — invoke dispatch decoupled from the GUI webview
- The backend is now generic over `tauri::Runtime` (pure signature threading, zero behavior change).
- `run()`'s wiring is extracted into `build_core_app<R>` with `AppMode { Gui, ServeHeadless }`: GUI-only plugins (single-instance, updater, dialog, clipboard) and window restore/close handling are gated; everything else — DB, state, PTY daemon, control socket, web-remote server, iroh, background loops — is shared.
- `build_headless_app()` boots the full app on Tauri's in-memory `MockRuntime` with **no display server**, creating an in-memory `main` webview so the existing zero-per-command `on_message` dispatch (all ~350 commands, ACL/arg-deserialization included) works unchanged headless.
- Headless hardening: dialog/clipboard commands return a clean error instead of hitting an unregistered plugin; native notifications are GUI-gated while the global `notification` event still emits for web clients.

### Phase 3 — `codemux serve`
- `codemux serve [--scope all|tailscale|loopback] [--port N] [--relay]`: refuses to start while a GUI (or another serve) holds the control endpoint — and the GUI reciprocally refuses while serve runs; binds through the same guarded enable path; prints reachable endpoints + a scannable pairing QR/link on startup; runs until SIGINT/SIGTERM with exit-cleanup parity.
- `codemux remote pair/enable/disable` work against a running `serve` from other SSH sessions.

## Guardrails (issue's regression list)

Auth and admission logic is structurally untouched — pairing single-use/TTL, session hashing, ticket-gated WS, origin checks, approval mode, revocation severing live sockets, account-mode admission (`load_cached_user`), and the iroh relay transport all ride the same unchanged code paths; they only gained a runtime type parameter.

## Verification

- `cargo test --lib`: **2230 passed, 0 failed** (same suite green before/after the genericity refactor — GUI behavior unchanged).
- New isolated integration tests (run with `DISPLAY`/`WAYLAND_DISPLAY` unset, temp HOME/XDG, PTY daemon disabled):
  - `serve_headless_dispatch` — headless boot + real invoke round-trip through the production dispatch path.
  - `serve_web_remote_roundtrip` — full `POST /api/pair` → `POST /api/ws-ticket` → `GET /ws?ticket=` → invoke → id-matched `ok` frame, zero GUI.
- Adversarial review pass over the diff: GUI wiring parity, mutual-exclusion race, serve idempotency (restore_on_boot + explicit enable), every `get_webview_window` call site classified for headless reachability, security invariants — no findings.
- Docs updated: `docs/features/web-remote-access.md` (headless server mode section), `docs/reference/CONTROL.md`, `docs/core/STATUS.md`, `docs/plans/web-remote-access.md`.

## Notes

- The embedded web UI ships in release builds; debug builds keep the existing Vite/`dist/` fallback (serve prints a warning if neither is present).
- Acceptable-intermediate per the issue: `serve` and the GUI are mutually exclusive on one machine (single-instance semantics preserved).
- Real-hardware acceptance (release build on a displayless box: `ssh box`, `codemux serve`, scan QR, drive terminal/agent from a phone) is the remaining manual smoke.


Zeus-Deus:
I fixed the failing Rust CI and two startup/shutdown correctness issues directly on this branch.

- The wiring contract tests were matching the old non-generic function signatures, so the runtime-generic command declarations looked absent. They now match stable function names.
- Headless startup could race `restore_on_boot`: the persisted `enabled` bit could make `codemux serve` report an already-running server and print pairing details before a listener existed (or hide a bind failure). Headless mode now leaves restoration to the awaited serve path, and `control_enable` treats the live runtime as authoritative.
- The success path used `process::exit(0)`, which skipped managed-state destructors. `run_serve` now returns a `Result`, performs cleanup, and drops the app normally; only `main` maps an error to exit code 1.

Local verification after merging current `main`: `cargo check`; 16 hook tests; 8 sandbox wiring-contract tests; `serve_headless_dispatch`; and the GUI-free pair/connect/invoke round-trip all pass. CI is rerunning now.

Zeus-Deus:
The rerun found one more CI-only test issue: the workflow builds `dist/` before `cargo test`, so a debug Tauri context may legitimately resolve `index.html`; the test had assumed every debug build used only Vite and required the asset to be absent. The resolver behavior was correct. The test now accepts both debug layouts and still requires embedded assets in release builds. The targeted headless dispatch test passes locally; CI has restarted.

PR #189, opened 2026-07-20
Render agent-produced images in chat tool results

Zeus-Deus:
## Problem

When an agent returns an image in a tool result — a screenshot from `Read` on a PNG, or any browser/screenshot tool — it never rendered in Agent Chat. Instead of an image, the transcript showed a wall of base64 text (or nothing).

Root cause is entirely frontend: the backend already forwards tool-result `content` to the client verbatim (it's opaque JSON on the Rust side, so the image block survives intact), but every tool body funnels `result_content` through a `contentToString` helper whose only special case is `{text}`. An image block `{type:"image", source:{type:"base64", media_type, data}}` fell through to `JSON.stringify`. Image rendering existed **only** for outgoing user attachments, never for agent-produced content.

## Fix (frontend-only)

- **`src/lib/agent-chat/tool-result-images.ts`** — `extractToolResultImages` normalises the two wire shapes (Anthropic base64 + `source.type:"url"`, and the OpenAI `image_url` variant) into a renderable `data:`/http `src`. Rejects non-image media types (e.g. PDFs share the base64 shape) and unsafe URL schemes (`javascript:`/`file:`).
- **`ToolResultImages.tsx`** — thumbnail row + click-to-open lightbox, mirroring the existing `UserMessage` attachment lightbox.
- **`ToolCallBodies.tsx` / `ToolCallBlock.tsx`** — `ToolCallBody` renders images below the per-tool body; both `contentToString` copies filter image blocks out (`isImageBlock`) so the payload never double-renders as base64 text, and `ToolCallBlock` renders nothing for an image-only result.
- **`ToolCallCard.tsx`** — auto-expands a result carrying an image, so a screenshot is visible without a manual expand.

Standalone assistant-authored image blocks (not wrapped in a tool result) remain unsupported — those would need a new backend `CompletedItem` variant.

## Verification

- New unit tests for the extractor (base64/url/image_url, media-type + URL-scheme rejection, non-array content).
- Full suite green: `npm run check` + `npm run test` (2934 passing).
- Visual check via the dev mock (seeded an image `Read` tool result in `richChatTurnEnvelopes`): the tool card auto-expands and renders the PNG inline; clicking opens the lightbox.

## Screenshots

Inline render (auto-expanded tool card) and lightbox verified against `npm run dev` with the seeded demo transcript.


Zeus-Deus:
I found two blockers in the current implementation:

1. `isImageBlock()` returns true for every `type: "image"` / `image_url` block, even when `imageFromEntry()` rejects it (for example `application/pdf`, a malformed source, or an unsafe URL). Both `contentToString()` paths then filter the block out, while `extractToolResultImages()` renders nothing, so the payload silently disappears. Please only suppress a block from the textual fallback when it is actually renderable (and add coverage for rejected blocks retaining a fallback).

2. `defaultExpanded` is only consumed by `useState()` on mount. Tool cards normally mount while `status === "running"` and receive `result_content` later, so a screenshot that arrives on completion does not auto-expand as described. Please react to the `hasImages` false→true transition (without repeatedly overriding a user collapse) and cover the running→image-result update.

The extraction/rendering direction and visual treatment otherwise fit the existing chat system well.

Zeus-Deus:
Implemented the requested corrections and re-reviewed the image path end-to-end. Raw payloads are now suppressed only for image blocks the renderer actually accepts, so malformed/PDF/unsafe-URL blocks remain visible. Cards also auto-open exactly once when an image arrives after mount and preserve a later manual collapse. Added helper, specialised/generic renderer, and async card-transition regressions (60 focused tests pass), and visually verified the seeded inline image in the dev mock.

PR #192, opened 2026-07-21
Fix Codex login status detection

Zeus-Deus:
## Summary

- classify Codex app-server authentication from both `account` and `requiresOpenaiAuth`
- keep logged-in ChatGPT/API-key accounts available in the model picker and session startup
- use the current `codex login status` CLI probe
- update the fake app-server, regression coverage, and canonical documentation

## Root cause

CodeMux treated `requiresOpenaiAuth: true` as a logged-out signal. Codex uses that field to describe whether the active provider requires OpenAI authentication, so it remains true for a valid logged-in account. Login is missing only when OpenAI auth is required and `account` is null.

## Verification

- `cargo check --manifest-path src-tauri/Cargo.toml`
- `CARGO_INCREMENTAL=0 cargo test --manifest-path src-tauri/Cargo.toml --test codex_adapter` — 35 passed
- `CARGO_INCREMENTAL=0 cargo run --manifest-path src-tauri/Cargo.toml --example codex_capabilities_smoke` — live account harvested 7 models
- `npm run check`
- `mise x node@22 -- npm run test` — 202 files / 2,924 tests passed
- `git diff --check origin/main...HEAD`


Zeus-Deus:
The conflict-updated branch exposed a Linux-only CI flake in the new capability-harvest test: under full-suite process load the temporary fixture wrapper can hit the same `ETXTBSY` (`Text file busy`) spawn race that `start_session_resilient` already handles. Windows and every other job passed.

I added an equivalent bounded retry for capability harvests, restricted to that exact transient error so auth/protocol failures still fail immediately, and made the unauthenticated assertion print unexpected results. Both capability-harvest tests pass locally; CI is rerunning.

PR #198, opened 2026-07-24
Sidebar Inbox: flat workspace inbox replacing the project tree

Zeus-Deus:
## What this is

The expanded left sidebar's nested project tree is replaced by a **flat workspace inbox** (from the `Sidebar Inbox.dc.html` design handoff): one card per active workspace, a project filter, and a persisted "Settled" section for swept-aside work. The collapsed rail and footer are rebuilt to match.

## Expanded sidebar

- **Header**: search field (opens the command palette, shows the resolved keybind) + neutral-ghost new-agent button; 32px controls
- **Project filter dropdown**: folder icon + current filter with rotating chevron, pinned dashed add-repository button; the panel lists All projects + each repo with its active-workspace count; filtering applies to both active cards and settled rows
- **Workspace cards**: repo eyebrow · work title (linked-issue title while the agent is live) + issue chip · needs-you blocker line · mono meta line (branch, ahead, +/- diff, PR chip, provider logos, remote cloud, notification badge) · agent state (working indicator / pulsing "Needs you" / "Done · review" / idle time) that swaps to a **Settle** button on hover/focus
- **Settle / un-settle**: cards collapse (~200ms) into a flat, recency-ordered Settled section of one-line rows (violet merge icon for merged PRs); Un-settle reverses. Visual only — nothing is archived or deleted. Persisted via UI state
- **Settle safety net**: working/blocked cards can't be settled; a settled workspace whose agent goes live auto-un-settles
- **Auto-settle**: fully idle cards settle themselves when their PR merges/closes, or after N idle days (Settings → Appearance → Sidebar, default 3d, off-able); manual un-settle sets a keep-active pin until the agent runs again
- **Settled-tail pagination**: 10 rows, then "Show 25 more (X hidden)"
- **Context menus**: both row shapes share the full workspace menu (rename/archive/delete/move-to-host) with Settle/Un-settle on top
- **Keyboard**: Alt+1–9 jumps to the Nth visible card (rebindable), with index badges while Alt is held

## Collapsed rail (52px)

New-agent + search on top, then one 28px avatar button per active workspace with its own status dot (red pulse / amber / green); click selects **without expanding**; settled workspaces never appear. Footer icons below.

## Footer nav (both states)

Automations + Workspaces as labeled ghost buttons, icon-only Ports (count badge) and the app-menu gear — fixed chrome, never in the scrolling list.

## Visual system

- Sidebar surface is now **darker than the main pane** (per-scheme `--sidebar` overrides in the custom token layer; survives preset re-apply)
- Cards are flat/transparent at rest; selection is lightness only (neutral border + lighter fill) — no accent-colored selection anywhere; needs-you red is the only colored border
- All colors are existing tokens (`status-*`, `accent-violet`); no hardcoded palette values
- Sidebar default width 256 → 288px

## Settings

- **Show git stats** (`sidebar.show_git_stats`, default on)
- **Auto-settle idle work** (`sidebar.auto_settle_days`, Off/1/3/7/14, default 3)
- Removed the dead `sidebar.live_agents` grouping setting

## Notes

- The superseded tree components (`sidebar-project-group`, needs-you strip, LIVE section, `SidebarWorkspaceRow`) remain in-repo but unmounted — `sidebar-workspace-row.tsx` still exports the context-menu/delete-dialog machinery the inbox reuses, and its test suite covers it. Removing the dead remainder is a follow-up.
- Card drag-reorder and the old hover flyout were dropped intentionally (cards keep stored order; the rail shows per-workspace status directly).

## Verification

- `npm run check` clean; **3,009 frontend tests** across 208 files passing (≈80 new: inbox store/persistence, settle lifecycle incl. safety net + auto-settle + pin, pagination, dropdown filter, jump targets/registry/dispatch, rail, footer)
- `cargo check` + `cargo test` passing (2,238; no Rust changes in this branch — the `git::tests::merge_into_base` family is flaky under full parallelism on the dev machine and passes with `RUST_TEST_THREADS=4`)
- Browser-driven visual QA on the dev mock at every step: inbox states, settle/un-settle round-trips, filter dropdown with counts, jump-hint badges, collapsed rail, footer, contrast inversion
- Docs updated: `docs/features/sidebar.md` (rewritten), `settings.md`, `SHORTCUTS.md`, `DESIGN-SYSTEM.md`, `INDEX.md`, `STATUS.md`


Zeus-Deus:
Reviewed the full diff (~3.6k lines, 35 files). The architecture is solid — the store is defensive (corrupt-JSON tolerant, prune-safe, write-throttled), context menus/archive/delete/push-to-host/rename are preserved on both row shapes, all colors go through tokens, no event-listener leaks, and the ~80 new tests cover the store, settle lifecycle, filter, pagination, jumps, rail, and footer. One real logic bug and two stale doc hunks should be fixed before merge:

**1. Keep-active pin is destroyed by merely selecting the workspace — un-settled merged-PR cards bounce straight back to Settled**

`sidebar-inbox.tsx` (activity-observation effect, ~line 367) calls `noteActivity(ws.workspace_id, at)` when `status !== null || isActiveWs || unseen`, and `noteActivity` (store, ~line 202) *always* clears the keep-active pin when set. So for the **active** workspace the pin is cleared even though its agent status is `null` — selection counts as "agent activity".

Concrete failure: click a settled merged-PR row (activates the workspace) → **Un-settle** → pin set → next render, the activity effect clears the pin (`isActiveWs`, status null) → the auto-settle effect sees `prDone && !pin && status === null` and re-settles it immediately. That violates the PR's own invariant ("keep-active pin suppresses auto-settle until its agent runs again" — stated in the code comments, settings copy, and sidebar.md).

The pin test ("keeps a merged-PR card active after the user un-settles it") runs with `activeWorkspaceId = ""`, so the `isActiveWs` path never fires — which is why the suite is green.

**Fix:** only clear the pin on genuine agent activity — e.g. give `noteActivity` a `clearPin` flag and pass `status !== null`, or split stamping (`isActiveWs`/`unseen`) from pin-clearing (`status !== null`) at the call site. Add a test where the un-settled workspace is the active one.

**2. Stale docs from earlier commits in this PR**
- `docs/INDEX.md` (sidebar entry) still says "search + repo filter chips" and "the rail keeps aggregate agent-status dots on project avatars + the per-project hover flyout" — the chips became a dropdown and the rail was rebuilt as a per-workspace strip (flyout / `sidebar-rail-projects.tsx` deleted) in later commits.
- `docs/reference/FEATURES.md` likewise still describes "Repo filter chips (All · one per repo · add-repo)".

**Follow-up-sized (fine to ride along or land separately):**
- `handleSettle` defers the store write 200ms behind the collapse animation and the unmount cleanup clears pending timeouts — collapsing the sidebar (Ctrl+B) inside that window silently discards the settle. Committing to the store immediately and keying the animation off a transient marker would close it.
- The collapsed rail dropped `notification_count` badges the old project avatars surfaced — if intentional, worth a line in sidebar.md's constraints.
- Space-key activation on cards/settled rows doesn't `preventDefault` (activates *and* scrolls); empty sticky-trigger label when the filtered project's last workspace vanishes; `prune()` only runs from the expanded inbox; `formatElapsed` shows "96h" instead of "4d".

Also heads-up: this overlaps with #203 on `src/components/settings/settings-view.tsx` — whichever merges second needs a manual rebase of that file. CI is green; merges clean against current main. Happy to merge once #1 and the doc hunks are fixed.

Zeus-Deus:
Confirmed all findings and fixed everything in d06dfa8 — including the follow-up-sized items, since they were all small enough to ride along.

**1. Keep-active pin (the real bug)** — verified exactly as described: the activity observer stamped the *active* workspace at status null and `noteActivity` unconditionally cleared the pin, so an un-settled merged-PR card bounced straight back to Settled on the next tick. Fixed with your suggested shape: `noteActivity(id, at, { clearPin })`, and the call site passes `clearPin: status !== null` — selection and first-seen baselines now only stamp; genuine agent activity is the only thing that un-pins. The invariant comments were updated to match. Added the regression test with the un-settled workspace as the **active** one (plus a case that a clearPin-less stamp leaves the pin) — and confirmed it fails against the pre-fix code. Good catch on why the old suite was green; the `activeWorkspaceId = ""` test is kept alongside.

**2. Stale docs** — INDEX.md and FEATURES.md now describe the project-filter dropdown (with counts) and the per-workspace rail; the chip/flyout wording is gone.

**Follow-ups, all taken:**
- Pending settle is now **flushed** on unmount (tracked in a ref; the cleanup commits it to the store instead of discarding) — collapsing the sidebar mid-animation no longer loses the settle. Test covers unmount-before-200ms with fake timers.
- Space-key activation `preventDefault`s on both row shapes.
- A filter pointing at a vanished project resets to All projects (test included).
- The rail now runs the same guarded `prune()` as the expanded inbox.
- `formatElapsed` gained a day tier (`4d` / `4d3h`).
- The rail's dropped notification badges are documented as intentional in sidebar.md's constraints.

Verification: tsc clean, full suite 3,034 tests / 209 files passing. Noted on the #203 overlap in `settings-view.tsx` — happy to rebase this side if that lands first.

PR #202, opened 2026-07-25
Polish agent chat composer and model picker

Zeus-Deus:
## Composer

- **Taller resting composer** — the idle card is now 140px (reserves ~4 input lines) and auto-grows to ~10 rows before scrolling internally, instead of a single-line strip.
- **Softer shell** — 20px radius, hairline border, deep low-opacity shadow.
- **Ghost footer controls** — the model / reasoning / speed / permission pills are now borderless text controls sharing one `FOOTER_TRIGGER` recipe, separated by 16px hairline pipes. Pipes render inside the capability-gated pickers, so a hidden control can never orphan a separator. Send/stop grew to 32px circles with a hover scale.
- **Tucked scope strip** — the Thread Scope controls render as a narrower bar attached flush under the card (inset by exactly the corner radius, bottom-only rounding, one tonal step less elevated). The running-thread Context Row uses the same strip shell, so the layered silhouette survives the first send.

## Model picker

- **Concrete Claude names** — alias rows (`default`/`opus`/`fable`/`sonnet`/`haiku`) now render the concrete model they resolve to ("Claude Opus 5", "Claude Fable 5"), parsed from the row's live version-bearing description so a CLI that re-points an alias stays correct without a catalog bump; qualifiers survive ("(1M context)", Recommended). Full-id rows with nickname labels promote the same way.
- **No duplicate default row** — the `default` alias folds into its concrete twin, which leads the list with a Recommended marker (row is kept when no twin exists, so nothing launchable is lost).
- **Compact 400px popup** — plus `minmax(0,1fr)` + `min-w-0` so the list column can actually shrink; previously its min-content forced a 455px column that overflowed the popup and clipped the provider rail.
- **Ctrl/Cmd+1..9 jump shortcuts** — bound to the first nine rows of the live filtered list (search, rail tab, and favorites order respected) via a window capture-phase listener, so they fire while the search box is focused; selection closes the picker. Per-row kbd chips show the platform-correct label.
- **Always-visible favorite star** and removal of the invisible trailing check-icon gutter every row was dragging (via the base CommandItem's `command-shortcut` slot).

Codex and OpenCode rows were already fine and pass through untouched.

## Testing

- `tsc --noEmit` clean; frontend suite green (2965 tests), including updated ComposerFooter / AgentChatPane / picker specs
- New Rust unit tests for the alias-name parser, alias promotion, default-row dedupe, and the no-twin fallback
- Visually verified in the dev mock at multiple viewports: composer geometry (140px resting card, flush 0px strip seam, 20px insets), centered 16px separator pipes, 400px popup with intact rail, chip/star flush right

### DIFF
--- a/docs/features/agent-chat.md
+++ b/docs/features/agent-chat.md
@@ -1563,8 +1563,12 @@ the "no panes" home landing lives in `ChatHomeLanding.tsx`.
 BOTH first-send surfaces — `DraftChatSurface` (lazy-creation draft) and
 `AgentChatPane`'s new-thread empty state (`messages.length === 0` on a
 real workspace's pane) — render their scope controls **below** the
-composer as one shared borderless `ThreadScopeRow`
-(`src/components/chat/pickers/ThreadScopeRow.tsx`), threaded through
+composer as one shared `ThreadScopeRow`
+(`src/components/chat/pickers/ThreadScopeRow.tsx`) — rendered as a
+**scope strip**: a narrower bar attached flush under the composer card
+(inset each side by the card's 20px corner radius, bottom-only
+rounding, one tonal step less elevated — the shared `SCOPE_STRIP` /
+`SCOPE_STRIP_INSET` class constants), threaded through
 `Composer`'s `belowComposerSlot` prop (a sibling of `zone1Override`,
 which both surfaces now pass `null` while a project root resolves — the
 cwd label moved into the row's location control). Three states, matched
@@ -1757,8 +1761,9 @@ title-bar chat tab). See `docs/features/gui-chrome.md`.
 
 Design: `.design-import/Context Row.dc.html` (claude.ai/design handoff).
 Once a thread has messages, `AgentChatPane`'s `belowComposerSlot` swaps
-`ThreadScopeRow` for a read-only **Context Row** — the same outer
-layout (`flex justify-between px-1`), but scope is committed (the first
+`ThreadScopeRow` for a read-only **Context Row** — the same tucked
+scope-strip shell (`SCOPE_STRIP` / `SCOPE_STRIP_INSET` exported from
+`ThreadScopeRow.tsx`), but scope is committed (the first
 send locked it in) so the left side is no longer clickable:
 
 - **Left** — static ghost-styled spans: folder icon (home icon +
--- a/src-tauri/src/agent_provider/claude/capabilities.rs
+++ b/src-tauri/src/agent_provider/claude/capabilities.rs
@@ -2006,34 +2006,35 @@ mod tests {
             sdk_model("haiku", "Haiku", &[], None, None),
         ];
         let caps = build_capabilities_from_sdk(live);
-        let desc = |id: &str| {
+        let row = |id: &str| {
             caps.models
                 .iter()
                 .find(|m| m.id == id)
                 .unwrap_or_else(|| panic!("{id} row must exist"))
-                .description
-                .clone()
-                .unwrap_or_else(|| panic!("{id} row must carry a description"))
         };
-        // `default` and `opus` both resolve to Opus 4.8, which defaults to
-        // the 1M context window → the "with 1M context" qualifier.
+        // The backfilled version now lives in the promoted LABEL; the
+        // description keeps only the blurb. `default` resolves to the
+        // same concrete model as `opus`, so it folds into that twin,
+        // which leads the list with the Recommended marker.
+        assert!(caps.models.iter().all(|m| m.id != "default"));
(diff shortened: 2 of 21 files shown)


Zeus-Deus:
Reviewed — the composer/strip layout work and the picker overflow/gutter fixes all check out (verified the mirror overlay tracking, separator gating, and the `command-shortcut` gutter reclaim). Two real defects in the model-roster logic should be fixed before merge, and the branch now needs a rebase:

**1. Persisted `model: "default"` ids dangle after the dedupe (moderate — hits every pre-existing Claude draft/thread)**

Before this PR the roster's first row was `default`, and `defaultModelId()` returns `models[0].id` — so existing persisted drafts (`chat-draft-store`) and thread records commonly store `"default"`. `dedupe_default_alias` now removes that id from the roster whenever a concrete twin exists (the common case), and nothing normalizes stored ids: `selectModel` returns `null`, so the footer trigger falls back to the raw string "default", the ReasoningPicker vanishes (`if (!model) return null`), `supports_images` flips off, and no picker row shows active. Sessions still launch, so it's a UI/state regression rather than a hard break — but it degrades every existing thread on upgrade.

**Fix:** on capability hydration, remap a stored `"default"` that's absent from the roster to `models[0].id` (or make the `selectModel` path fall back to `models[0]` for `"default"` specifically).

**2. `dedupe_default_alias` runs before `append_missing_maintained_families`, recreating the duplicate it exists to remove (edge case)**

`build_capabilities_from_sdk` calls `dedupe_default_alias` *then* `append_missing_maintained_families` (capabilities.rs:807-808). The append exists precisely for the roster shape where the CLI demoted a family into its `default` alias — i.e. `default` present but no concrete twin. In that case dedupe finds no twin and keeps the promoted `default` row (say "Claude Opus 4.8"), then the append adds maintained `claude-opus-4-8` with the same label — exactly the duplicate-concrete-name pair dedupe was written to eliminate. The new tests miss it because their no-twin fixture resolves `default` to a different family than the maintained entry.

**Fix:** swap the order — run `dedupe_default_alias` after `append_missing_maintained_families`. The new tests still pass under that ordering.

**Minor (non-blocking):**
- `Composer.tsx` hardcodes the shadow color `rgba(0,0,0,0.45)`; `shadow-[0_12px_28px_-18px] shadow-black/45` would at least separate geometry from color per the design-system rule. Also 0.45 alpha is much heavier than the `shadow-sm` it replaces — worth a light-theme visual check.
- The Ctrl+1..9 jump shortcuts match on `event.key`, so on layouts where digits are shifted (e.g. AZERTY) the advertised chips are unreachable (`shiftKey` is rejected, and unshifted produces a non-digit key). Matching `event.code === "Digit1".."Digit9"` / `Numpad1..9` fixes it.

**Rebase needed:** #200 just merged and deleted the scope-hint `<p>` block in `ThreadScopeRow.tsx` that this branch edits (`px-4`), plus its hint tests — the coherent resolution is this PR's `SCOPE_STRIP` shell *without* the hint block. CI is green on the current head; happy to merge once #1/#2 are addressed and the branch is rebased.

Zeus-Deus:
All four points were valid — addressed in bc82da9 (branch also rebased onto main per the note):

1. **Stale `"default"` ids** — `selectModel` now falls back to `models[0]` for the literal `"default"` when it's absent from the roster (doc comment explains the persisted-alias history). The trigger label/subtitle/tooltip, `modelLabel`, and the active-row highlight all route through it, so pre-existing drafts/threads keep their reasoning/speed pickers and highlight the Recommended row. Other unknown ids still resolve to null — no over-broad fallback. Covered by 13 new tests across the store, capability-defaults, and the picker.

2. **Ordering** — `append_missing_maintained_families` now runs before `dedupe_default_alias`, with comments on both sides explaining why. The appends-opus test was reworked into the regression test for your exact scenario: `default` + no concrete Opus row → single folded `claude-opus-4-8` leading the list with the Recommended marker, plus an explicit duplicate-count assertion.

3. **Shortcuts** — now matched on `event.code` (`Digit1..9` / `Numpad1..9`), so shifted-digit layouts reach them; modifier guards unchanged. Tests cover a shifted-layout keypress (`key: "&", code: "Digit1"`) and numpad.

4. **Shadow** — split into `shadow-[0_12px_28px_-18px] shadow-black/40` (geometry/color separated, one step lighter); verified it composes to `oklab(0 0 0 / 0.4) 0 12px 28px -18px` at runtime.

Rebase resolution matched your suggestion: the `SCOPE_STRIP` shell without the removed hint block (stale doc comment cleaned up too). Full suites green locally: 2978 frontend tests, full Rust lib, tsc.

PR #203, opened 2026-07-25
Add user-configurable default viewport for agent browser sessions

Zeus-Deus:
## Motivation

The agent browser's built-in 1280×800 baseline reads as "zoomed in" on larger monitors — screenshots the agent takes while testing issues/PRs show the narrow-desktop responsive layout instead of what the user sees in their own browser. This adds a synced setting so the default matches the user's screen (e.g. QHD 2560×1440 for a 27" 1440p display).

## What's new

**Settings → Browser → "Default viewport"** (`UserSettings.browser.default_viewport`, synced): Default (1280×800) / Full HD (1920×1080) / QHD (2560×1440) / 4K (3840×2160). The stored value is any `browser_viewport::parse_spec` string (preset name or `WxH`).

Consumption points:

- **Fresh daemon launch** (`AgentBrowserManager::run_command`): when a session's `running` flag flips false→true, the configured viewport is applied via `set viewport` *before* the requested action — so even a first-command screenshot renders at the user's size. Fires once per daemon lifetime, only when the setting is present; skipped for explicit `viewport` and `close` actions, and never re-applied to a running daemon, so an agent's own viewport choices (e.g. mobile testing) are never stomped.
- **`viewport reset`** (CLI `codemux browser viewport reset` + MCP `browser_viewport {preset:"reset"}`): resolves through new `parse_spec_configured` → lands on the configured default instead of the hard-coded `RESET_SPEC`. Both `viewport-presets` listings report the actual reset target.
- **Peek overlay** (`BrowserPeekOverlay.tsx`): the "Desktop-size background browser" pin now uses the configured default (`parseViewportString`, 1280×800 fallback), keeping the popup consistent with what screenshots show.

## Safety

- Unset (the default) keeps prior behavior byte-for-byte.
- Invalid synced values (garbage from another device) degrade to the 1280×800 baseline instead of erroring — a bad blob can never break browser startup.
- Old settings blobs deserialize via serde defaults (back-compat tested).

## Verification

- `cargo check` clean; lib tests: browser_viewport 30/30, settings_sync 20/20, agent_browser 61/61
- `tsc --noEmit` clean; frontend suite 204 files / 2969 tests passed
- Live UI check in the dev mock: the select renders, opens, and persists the QHD selection; `codemux browser viewport reset` returns 1280×800 with no setting configured

Docs: new "User default viewport" section in `docs/features/browser.md`; schema + UI table updated in `docs/features/settings-sync.md`.

### DIFF
--- a/docs/features/browser.md
+++ b/docs/features/browser.md
@@ -171,7 +171,46 @@ existing frame-draw letterbox scales the full-size frame down to fit. Input
 mapping needs no changes — `mapToViewport` already routes clicks through the
 canvas rect + draw rect into viewport coordinates. Applies only to the peek
 overlay; normal browser panes (and the peek with the setting toggled off) keep
-the container-sync behavior unchanged.
+the container-sync behavior unchanged. When a user default viewport is
+configured (next section), the peek pins to that instead of 1280×800 —
+`DESKTOP_PEEK_VIEWPORT` is only the fallback.
+
+## User default viewport (`browser.default_viewport`)
+
+Settings → Browser → "Default viewport" (`UserSettings.browser
+.default_viewport`, synced settings, default unset). Motivation: the built-in
+1280×800 baseline reads as "zoomed in" to users on larger monitors — agent
+screenshots for PRs/issues show the narrow-desktop responsive layout instead
+of what the user sees in their own browser. Setting e.g. QHD (2560×1440) makes
+agent screenshots match the user's screen proportions. Cosmetic cost only:
+frames/screenshots scale ~linearly with pixel count, and model-side vision
+downscales past ~1.15 MP anyway, so agent token cost is unchanged.
+
+The value is a `browser_viewport::parse_spec` string — preset name or `WxH`.
+Consumption points (all lenient — an invalid synced value degrades to the
+1280×800 baseline, never errors):
+
+- **Fresh daemon launch** (`AgentBrowserManager::run_command`): when the
+  session's `running` flag flips false→true, the configured viewport is
+  applied via a `set viewport` command BEFORE the requested action, so even a
+  first-command screenshot renders at the user's size. Fires once per daemon
+  lifetime and only when the setting is present; skipped for `viewport`
+  (explicit size wins) and `close` (don't boot a daemon to resize it). An
+  agent's own later `viewport` calls are never stomped.
+- **`viewport reset`** (CLI `codemux browser viewport reset` + MCP
+  `browser_viewport {preset: "reset"}`): resolves through
+  `browser_viewport::parse_spec_configured`, which lands on the configured
+  default (`configured_default_spec`) instead of the hard-coded `RESET_SPEC`.
+  Both `viewport-presets` listings report the actual reset target. The CLI is
+  the same binary as the app, so it reads the same on-disk settings cache.
+- **Peek overlay pin** (`BrowserPeekOverlay.tsx`): with "Desktop-size
+  background browser" ON, the `fixedViewport` prop uses
+  `parseViewportString(browser.default_viewport)` (WxH strings only on the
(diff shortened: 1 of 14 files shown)


Zeus-Deus:
Reviewed — the wiring is solid overall (settings sync, serde back-compat, CLI/MCP reset resolution, peek overlay pinning all check out), but there's one real functional gap that should be fixed before merge:

**1. Default viewport is never re-applied after an agent closes and reopens the browser** (`src-tauri/src/agent_browser.rs`, the new `was_running` block in `run_command`)

The agent-facing `close` action is routed through `run_command` via the Tier-1 `_` match arm in `control.rs` — `AgentBrowserManager::close()` (which removes the session entry) is never called on this path. `run_command` unconditionally does `std::mem::replace(&mut s.running, true)` for **every** action, including `close`, and nothing resets the flag after the daemon is killed. So:

```
browser open   → was_running=false → viewport applied ✓
browser close  → running stays true, daemon killed
browser open   → fresh daemon, but was_running=true → viewport silently skipped ✗
```

That's a normal per-task agent pattern, and it contradicts the "fires once per daemon lifetime" intent — the guard actually fires once per session-map-entry lifetime, which outlives the daemon across close/reopen. (Only `start_stream`'s dead-socket probe ever clears the flag.)

**Fix:** after a successful `close` in `run_command`, reset the session state (`s.running = false; s.pid = None;`) — or don't set `running = true` for the `close` action in the first place. Either restores the false→true transition for the next launch. A regression test for the open→close→open sequence would lock this in.

**2. Minor: `resolve_default` is production-dead** (`src-tauri/src/browser_viewport.rs`)

`resolve_default` is only called from the new unit tests; the production path (`configured_default_viewport`) re-implements the same trim/empty/`"reset"`-rejection/parse-fallback logic by hand. The tests validate a copy of the logic, not the code that ships, so the two can drift. Suggest having `configured_default_viewport` delegate to `resolve_default` (or testing through `configured_default_*` directly).

CI is green; happy to merge once #1 is addressed.

Zeus-Deus:
Both points were valid — fixed in 0e8006a.

**1. Close/reopen gap — confirmed and fixed.** Verified the routing independently: the agent-facing `close` indeed runs through `run_command`'s Tier-1 arm, and `AgentBrowserManager::close()` is only reached via workspace teardown, so the entry kept `running = true` across the daemon's death. The state handling is now split into two helpers that `run_command` itself calls:

- `note_action_start` — the `was_running` capture; for `close` it only *reads* the flag (the early-mark exists to protect a daemon from a concurrent `start_stream`, which is pointless when we're about to kill it).
- `note_close_finished` — resets `running = false; pid = None` after the close executes, keeping the port/map entry intact. It runs unconditionally rather than only on `Ok`: post-close the daemon state is unknown at worst, and both consumers of `running` tolerate a stale `false` (`start_stream` re-probes the socket; a redundant viewport apply on a live daemon is harmless right after the agent chose to close it). Gating on success would leave the original bug in place for the close-errors-on-dead-daemon edge.

Regression test `close_resets_running_so_reopen_reapplies_default_viewport` locks in the open→close→open transitions through those same helpers (no shelling out to the real binary): open→false, open→true, close→true (twice, proving close doesn't mutate), reset→entry survives with `running=false`/`pid=None`, open→false again.

**2. `resolve_default` drift risk — fixed.** Added a single pure core `resolve_default_setting(Option<&str>) -> Option<ViewportSpec>`; `configured_default_viewport()` now delegates to it (hand-rolled duplicate removed) and `resolve_default()` is just `.unwrap_or(RESET_SPEC)` on top. New test covers the Option-returning path directly, so the unit tests now validate the code that ships.

`docs/features/browser.md` updated with the close-reset semantics. Verification: `cargo check` clean; agent_browser 62/62, browser_viewport 31/31, settings_sync 20/20.

PR #204, opened 2026-07-26
feat(chat): show real image thumbnails in attachment chips

Zeus-Deus:
## Problem

Every pasted image staged the same generic image icon behind the same `pasted-image.png` label, so two attachments in the composer strip were impossible to tell apart.

## Change

Image chips now render the actual image:

- **30×22px thumbnail** in the chip's icon slot, built from the attachment's in-memory `resolvedImage.bytes` — already retained for the optimistic user bubble, so this costs no extra IPC or disk read.
- A chip with a preview swaps the full pill radius for `rounded-md` with a 3px inset. A 22px-tall thumbnail inside a `rounded-full` pill pokes through the corner arc, and a rounded rect reads as "picture" anyway. Chip height goes 24px → 28px; non-image chips are byte-identical.
- **Hover tooltip gains a ≤260×180px preview** above the existing filename / token-estimate / size lines. 30px is enough to tell two screenshots apart by colour; the tooltip is where you actually read one.

## Robustness

- Object URL is revoked on unmount, so removing a chip doesn't leak it.
- Unresolved bytes (still staging), a decode failure, or an environment without `URL.createObjectURL` (jsdom under vitest) all fall back to the original icon pill.

## Verification

- `npm run check` clean; 3056 frontend tests pass across 209 files.
- 5 new `AttachmentChip` tests: thumbnail render, chip shape swap, URL revocation on unmount, decode-failure fallback, loading state.
- Visually verified in the dev-mock UI with two pasted images (chip strip + hover preview).

`npm run verify` can't complete its Rust half in a fresh worktree — `src-tauri/binaries/agent-browser-x86_64-unknown-linux-gnu` is absent, which is a worktree bootstrap gap unrelated to this frontend-only change.

## Docs

`docs/features/agent-chat.md` — new bullet under Attachments.

### DIFF
--- a/docs/features/agent-chat.md
+++ b/docs/features/agent-chat.md
@@ -161,6 +161,18 @@ The chat pane stack:
   images via paste / drop / picker. Inline chips, send-time injection,
   expand, caps, gif guard, chip tooltips. See
   `docs/research/step-8-attachments.md`.
+  - **Image chips show the image, not a generic icon.** `AttachmentChip`
+    builds an object URL from the attachment's in-memory
+    `resolvedImage.bytes` (already kept for the optimistic user bubble,
+    so this costs no IPC) and renders a 30×22px thumbnail in the icon
+    slot; the chip swaps from the pill radius to `rounded-md` with a 3px
+    inset so the preview reads as a picture, and the hover tooltip adds a
+    ≤260×180px preview above the filename/token lines. Two pasted
+    screenshots were previously indistinguishable — both chips read
+    `pasted-image.png` behind the same icon. The URL is revoked on
+    unmount; missing bytes, an image that fails to decode, or an
+    environment without `URL.createObjectURL` (jsdom) all fall back to
+    the original icon pill.
   - **Ctrl+V image paste has a Linux clipboard fallback.** WebKit2GTK
     strips image payloads from the JS `paste` event, so `Composer`'s
     handler tries `e.clipboardData.items` first (works in the browser
--- a/src/components/chat/AttachmentChip.test.tsx
+++ b/src/components/chat/AttachmentChip.test.tsx
@@ -237,6 +237,95 @@ describe("AttachmentChip", () => {
     });
   });
 
+  describe("image thumbnail preview", () => {
+    /** jsdom implements neither half of the object-URL API, so patch both
+     *  onto the real `URL` and assert the created URL is used and revoked. */
+    function stubObjectUrls() {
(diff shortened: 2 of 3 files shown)


Zeus-Deus:

The core of this is correct: object URL created from bytes already in memory, revoked on unmount and on bytes change, `metadata.isLoading` correctly taking precedence over `showPreview`, decode failure degrading to the icon pill rather than a broken-image glyph, and the `typeof URL.createObjectURL !== "function"` guard covering jsdom. No IPC added. `data-preview={showPreview || undefined}` renders as `data-preview="true"`, matching the test.

Three things I'd tighten — none of them blocking:

### 1. `previewFailed` is never reset

```ts
const [previewFailed, setPreviewFailed] = useState(false);
```

It latches for the life of the component. If `attachment.resolvedImage` is ever replaced (a re-resolve, a retry, a staging path that lands bytes in two steps), the chip stays on the icon even though the new bytes might decode fine. Cheap fix: reset it in the same effect that builds the URL, or key it off `image`.

### 2. The effect keys on object identity, not content

```ts
useEffect(() => { ... }, [image]);
```

`image` is `attachment.resolvedImage`. If anything upstream ever reconstructs the attachment object with an equivalent `resolvedImage`, this revokes and recreates the URL — which means the `` changes and the thumbnail visibly re-decodes. Worth confirming the store hands back a stable `resolvedImage` reference; if not, `[image?.bytes, image?.mime]` is tighter.

### 3. The chip row has no `items-center`

The container in `Composer.tsx` is `flex flex-wrap gap-1.5 px-3 pt-2` — no `items-center`, so `align-items: stretch` applies. Growing the image chip from 24px to 28px therefore stretches its siblings (the Continue-run chip, `ModePill`, and a PR chip) to 28px too whenever an image is staged alongside them. Their contents stay centred because each is `inline-flex items-center`, so this may well look fine — but it's a visual change to chips this PR describes as "byte-identical", and it's worth an eyeball with an image + PR chip staged together.

Verified separately: `AttachmentChip` has exactly one consumer (`Composer.tsx:2158`), so there's no other surface where the height change lands.

Zeus-Deus:

I traced the `resolvedImage` → `Blob` → object-URL lifecycle, cleanup on replacement/unmount, decode-error fallback, chip/tooltip rendering, loading behavior, and accessibility impact. I also ran the focused `AttachmentChip` suite: 32/32 tests passed.

GPT 5.6 Sol xhigh found no errors in this PR.

PR #208, opened 2026-07-26
feat(sidebar): hover details card for workspaces

Zeus-Deus:
## Why

Every sidebar surface is lossy. An inbox card truncates its title and never renders `git_behind` or `git_changed_files` at all — both already ride on `WorkspaceSnapshot` and were displayed nowhere. A settled row shows only a title. A collapsed rail avatar shows nothing but a status dot, behind a title-only tooltip.

Resting the pointer on any of the three now opens a shared read-only `HoverCard` to the right (350ms open, 290px, `side="right" align="start"`).

## What it shows

Header — repo eyebrow, provider marks, agent state with elapsed, the **full untruncated title**, and the linked issue's number + title.

Rows, all conditional except Location — Branch, Uncommitted `+A −D`, Changed files, Ahead, Behind (or a single "Working tree · clean" row when there is nothing to report), Pull request, Issue, Port(s) capped at 3 with a `+N` overflow, Location (`This device` / host name / `host · in place`), Notifications muted.

Footer — the real path on disk (`worktree_path ?? remote_cwd ?? cwd`), `$HOME` collapsed to `~`, wrapped rather than truncated.

**No new Tauri command.** Every field is already on `WorkspaceSnapshot` or in `appState.detected_ports`. Styling reuses the `DetailRow` label/value shape from the context bar's popover rather than inventing a second vocabulary.

## Implementation notes

- The body is a separate component so Radix's unmount-when-closed keeps a sidebar of N workspaces from subscribing N times to the ports/hosts stores at rest.
- Ports are selected as a raw slice and narrowed in a `useMemo`. Filtering inside the zustand selector returns a fresh array per call, which breaks `useSyncExternalStore` snapshot equality and spins React into an infinite re-render loop — this crashed the card tree during development and was caught only by running the real app.
- Remote-ness keys off `host_id` alone. The hosts list loads asynchronously, so an unresolved name falls back to "Another device" rather than letting a failed lookup claim the workspace is local.
- Mounted inside `WorkspaceInboxMenu` on a different node than `ContextMenuTrigger`, so the two `asChild` triggers never compose onto one element.
- The label holds its width and the value truncates. The reverse rendered long worktree branches as `Bran…` next to the thing you came to see.
- The card is `pointer-events-none` — a detail surface, not an interactive one. Reviewers may want to overrule this: it means you cannot mouse in to select the branch name or path.

Removes the rail's title-only tooltip and the card's redundant native `title` attribute, which would have raced the hover card on the same element.

## Verification

- `npm run check` clean.
- `npm run test` — 210 files, 3069 tests pass, including 19 new ones covering the git rows, clean-tree fallback, non-git workspaces, PR/issue tones, port capping, the async-host fallback, mute, and path shortening.
- Visually confirmed in the dev build across all three surfaces (inbox card, settled row, collapsed rail), with instrumented `pointerenter` proving real mouse input opens it.
- Zero raw Tailwind palette utilities — status/accent tokens only, per the design system.
- `npm run verify` does **not** complete in this worktree: `cargo check` fails on a missing `src-tauri/binaries/agent-browser` sidecar. Pre-existing environment gap, unrelated — this branch touches no Rust.

Docs updated in `docs/features/sidebar.md` (inbox section, rail section, What Works, Constraints, Touch Points).

### DIFF
--- a/docs/features/sidebar.md
+++ b/docs/features/sidebar.md
@@ -120,12 +120,12 @@ visual only — nothing is archived, closed, or deleted.
   archive, delete) shared with the old row, including the delete/push-confirm
   dialogs.
 - **Hover details** (`workspace-hover-card.tsx`): resting the pointer on any
-  workspace surface — an active card, a settled one-line row, or a collapsed
-  rail avatar — opens a shared read-only `HoverCard` to the right (350ms open,
-  120ms close, `side="right" align="start"`, 290px). It exists because every
-  sidebar surface is lossy: the card truncates its title and drops
-  `git_behind` / `git_changed_files` entirely, the settled row shows only a
-  title, and the rail shows nothing but an avatar. Contents, all conditional
+  workspace surface — an active card, a settled or snoozed one-line row, or a
+  collapsed rail avatar — opens a shared read-only `HoverCard` to the right
+  (350ms open, 120ms close, `side="right" align="start"`, 290px). It exists
+  because every sidebar surface is lossy: the card truncates its title and
+  drops `git_behind` / `git_changed_files` entirely, the settled and snoozed
+  rows show only a title, and the rail shows nothing but an avatar. Contents, all conditional
   except the header and Location: repo eyebrow + provider marks + agent state
   (with client-derived elapsed), the **full** untruncated title, the linked
   issue's number + title, then label/value rows — Branch, Uncommitted `+A −D`,
@@ -494,8 +494,9 @@ Replaced the old project-avatar rail (aggregate dots + hover flyout,
 - Rail: one avatar button per active (neither settled nor snoozed) workspace,
   with individual status dots, select-without-expand, and the shared footer
   destinations.
-- Hover details on every workspace surface (card, settled row, rail avatar) —
-  full title, complete git picture, PR/issue, ports, device, and path on disk.
+- Hover details on every workspace surface (card, settled row, snoozed row,
+  rail avatar) — full title, complete git picture, PR/issue, ports, device,
+  and path on disk.
 - Per-workspace agent status covers both terminal and Agent Chat (Beta) agents
   (chat sessions publish into the same `pane_statuses` snapshot).
 - The done-review checkmark **survives an app restart**. `save_persisted_state`
@@ -530,7 +531,7 @@ Replaced the old project-avatar rail (aggregate dots + hover flyout,
   card. The backend now persists `last_active_at`, which would fix this; the
   card just doesn't read it yet (the settled shelf does, via `workEndedAt`).
   The hover card's elapsed reading inherits exactly this limitation (it reads
-  the same store, sampled once at open rather than off a ticking clock).
+  the same store, ticking on the shared coarse clock while the card is open).
 - Shelf collapse state (Settled open, Snoozed closed) and the settled paging
   window are **component state** — not persisted, and reset on every mount.
 - Multi-selection is session-only and cleared by a plain click, an activation,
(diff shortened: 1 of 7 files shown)


Zeus-Deus:

Solid, and the two hardest details are handled correctly:

- **The ports-selector trap is genuinely avoided.** Selecting the raw slice and narrowing in `useMemo` is the right shape; a filtering selector really would break `useSyncExternalStore` snapshot equality and spin. Good that this is written down in the code, not just the PR body.
- **The two `asChild` triggers are on distinct nodes** in all three call sites, which is the failure mode I'd expect from this composition.
- Remote-ness keying off `host_id` alone with an "Another device" fallback is the right call given the async hosts load.

### `pointer-events-none` contradicts the footer's own rationale

```tsx
{/* Real path on disk, last: the least-scannable line, and the one users
    most often want to copy or confirm. Wraps rather than truncates. */}
```

…inside a card that is `pointer-events-none`, so it cannot be selected or copied. You already flagged this for reviewers, so: I'd overrule it. The path and the branch name are exactly the two values a user reaches for a hover card to *get*, not just to read. Radix `HoverCard` handles mouse-into-content fine (that's what `closeDelay` is for), and dropping the class costs nothing except that the card intercepts clicks in its own footprint — which is over the main pane, not over other sidebar rows. If it stays non-interactive, the footer comment should stop claiming users want to copy it.

### Settled rows always report "Idle"

`SettledRow` passes `status={null}` unconditionally, so the header renders `Idle` even for a settled workspace whose agent is actively working — and #206 makes "settled but working" a real, reachable state (its auto-un-settle fires on `working`). The row itself shows no status either, so this isn't a regression, but the hover card is the surface whose whole purpose is to un-lose what the row dropped. `getWorkspaceStatus(workspace.surfaces, paneStatuses)` is already available.

### Two smaller notes

- **`elapsed` freezes at open.** `formatElapsed(Date.now() - stateSince)` is computed once per mount, deliberately. Fine for a 350ms-to-open card you glance at — but it's the one value on the card that visibly goes wrong if you leave the pointer resting, and "Working 4m" that stays 4m for two minutes reads as a bug.
- **The bare `<div>` wrapper in `SettledRow`** now sits between the settled list's flex container and the row, so it becomes the flex item. Almost certainly fine, but it's an unlabelled layout node — worth a `contents`-style comment or `className="contents"` if the row was relying on being a direct child.

Zeus-Deus:
Review by GPT 5.6 Sol xhigh found a path-boundary bug.

`shortenPath` at `workspace-hover-card.tsx:48-50` uses raw `path.startsWith(homeDir)`. That treats sibling prefixes as descendants: with home `/home/u`, a workspace at `/home/u2/project` is rendered as `~2/project`. The same issue occurs with trailing separators and platform-specific roots.

Please compare path components (or require the next character to be a separator, while handling exact-home as a special case) and add the sibling-prefix regression. The focused hover-card suite passed 19/19 tests; it currently covers an actual child and a completely unrelated path, but not this boundary.

Coordination note: this head has pairwise textual conflicts with #206 and #213.

Zeus-Deus:
Verified: all five review issues are genuinely fixed at the head (pointer-events restored on the card, settled rows pass real status, elapsed ticks via the shared coarse clock, `shortenPath` sibling-prefix/exact-home/trailing-separator cases covered). `tsc` clean and all touched suites pass (23/23 hover-card, 30/30 sidebar-inbox, 6/6 rail).

Blocked on a rebase now that #206 merged — conflicts in `sidebar-inbox.tsx`, `sidebar-inbox.test.tsx`, `sidebar-inbox-card.tsx`, `sidebar-rail-workspaces.tsx`, `docs/features/sidebar.md`. Should be mostly textual; one semantic point: main's settling work adds new tiers/states for settled rows, and this PR's settled-row status prop should reflect those (the hover card anticipated exactly this, so integration should be natural).

PR #214, opened 2026-07-27
fix: persist done-review checkmarks across app restarts

Zeus-Deus:
## Problem

The sidebar's "Done · review" checkmark did not survive a restart. Every finished workspace came back looking already-reviewed, which makes the inbox much less useful — the whole point is to show what you have not looked at yet.

`save_persisted_state` cleared the entire `pane_statuses` map before every write:

```rust
snapshot.pane_statuses.clear();
```

So `layout.json` always hit disk with `pane_statuses: {}`, and boot (`lib.rs:439-455`) just deserializes that file — there is no re-derivation. Every workspace initialized to idle.

This is not a regression from the sidebar rework. The `clear()` predates it: it came in with `70e97f98` (March) and is the only commit that ever touched that line. The rework commits (`6cf275ef`, `a9d61bb8`, `8fe9ec00`, `d06dfa89`) are all later and never went near status persistence.

## Fix

Replace the wholesale clear with `retain_persistable_pane_statuses`, which keeps `Review` and drops the rest:

```rust
snapshot.pane_statuses.retain(|pane_id, status| {
    *status == PaneStatus::Review && live_panes.contains(pane_id)
});
```

Three notes on the shape:

- **`Working`/`Permission` are still cleared.** The original comment was correct about those — the agent is dead after a quit, so restoring them would render a spinner or permission badge that no process backs. `Review` is a durable fact, not live process state.
- **Orphan pruning is load-bearing, not defensive.** Nothing removes a status when its pane disappears (`close_pane` leaves the key behind; the openflow/browser strips delete panes wholesale). That was harmless while the map was wiped every save, but would grow without bound now that it persists. The filter runs *after* those strips, so stale entries are cleaned up on the next write.
- **Nothing wipes the restored badge at boot.** All seven `set_pane_status` call sites are event-driven (`agent_chat.rs:360` is the close-pane handler, not a boot path), and `clear_transient_pane_status_by_session` already spares `Review` deliberately. Badges clear the usual way, via `activate_workspace`.

## Tests

Two regression tests in `state_impl.rs`:

- `review_status_survives_persistence_but_live_process_statuses_do_not`
- `persisted_pane_statuses_drop_entries_whose_pane_is_gone`

## Verification

`cargo check`, `tsc --noEmit`, and `npm run test` (209 files / 3051 tests) all clean; both new tests pass.

Two unrelated things surfaced while running the Rust suite, neither caused by this change:

- 10 `terminal::tests::process_kill` failures appeared only when the suite was run concurrently with `npm test` — those tests call `openpty` and were losing the race for resources. Run serially, all 13 pass.
- 2 `agent_provider::opencode` *live-binary* tests fail; reproduced on a clean tree with this change stashed, where a different one of them failed. They shell out to a real `opencode` binary and are environment-dependent and flaky.

## Docs

`docs/features/sidebar.md` updated — the checkmark's restart behavior is now documented under "What Works Today".

### DIFF
--- a/docs/features/sidebar.md
+++ b/docs/features/sidebar.md
@@ -183,6 +183,13 @@ Replaced the old project-avatar rail (aggregate dots + hover flyout,
   select-without-expand, and the shared footer destinations.
 - Per-workspace agent status covers both terminal and Agent Chat (Beta) agents
   (chat sessions publish into the same `pane_statuses` snapshot).
+- The done-review checkmark **survives an app restart**. `save_persisted_state`
+  keeps `PaneStatus::Review` entries in `layout.json` and drops
+  `Working`/`Permission` (dead processes after a quit) plus entries whose pane
+  no longer exists — see `retain_persistable_pane_statuses` in
+  `src-tauri/src/state/state_impl.rs`. Previously the whole map was cleared on
+  every save, so finished workspaces came back looking already-reviewed. The
+  badge still clears the normal way, on activating the workspace/tab.
 
 ## Current Constraints
 
--- a/src-tauri/src/state/state_impl.rs
+++ b/src-tauri/src/state/state_impl.rs
@@ -4480,6 +4480,37 @@ fn persisted_layout_path() -> Option<PathBuf> {
     Some(base.join(crate::APP_DIR_NAME).join("layout.json"))
 }
 
+/// Drop the pane statuses that mean nothing after a restart, keeping the ones
+/// that do.
+///
+/// `Working` and `Permission` describe a live process. The agent is dead once
+/// the app exits, so restoring a spinner or a permission badge would show the
+/// user a state that no process backs — the reason this whole map used to be
+/// cleared wholesale.
+///
+/// `Review` is a different kind of fact: it means "this agent finished and you
+/// have not looked at the result yet", which is just as true after a restart as
+/// it was before. Persisting it is what keeps the sidebar's done-checkmark
+/// alive across a quit; clearing it made every finished workspace come back
+/// looking like the user had already reviewed it.
+///
+/// Entries whose pane no longer exists are dropped too. Nothing removes a
+/// status when its pane goes away (`close_pane` leaves the key behind, and the
+/// openflow/browser strips above delete panes wholesale), so without this the
+/// map would grow forever now that it survives restarts.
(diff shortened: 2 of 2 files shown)


Zeus-Deus:

**No errors found.** I traced the claims in the description against the code rather than taking them at face value, and each one holds up:

- **Ordering is right.** `retain_persistable_pane_statuses` runs *after* `strip_openflow_panes_from_snapshot` / `strip_browser_panes_from_snapshot`, so panes those strips delete have their statuses pruned in the same pass. The "orphan pruning is load-bearing" note is accurate.
- **`Review` really is the complete keep-set.** `PaneStatus` is exactly `Idle | Working | Permission | Review`, and the field doc already states only non-idle entries are stored, so `*status == PaneStatus::Review` leaves nothing behind.
- **The filter can't corrupt live state.** `save_persisted_state` operates on an owned copy (`snapshot = strip_browser_panes_from_snapshot(snapshot)`), so the retain only shapes what hits `layout.json` — in-memory statuses are untouched.
- **Archived workspaces are a non-issue.** I checked this specifically, because a filter that walks only `snapshot.workspaces` while `archived_workspaces` is a separate `Vec` looked like it could silently drop badges. It can't: `ArchivedWorkspaceSnapshot` is metadata-only (`archive_id`, `title`, `cwd`, `worktree_path`, …) with no `surfaces`, and a restore mints a brand-new workspace — so an archived entry has no pane ids to lose.
- Both new tests genuinely pin the behaviour they claim to (the orphan test would fail if the `live_panes` check were dropped; the first would fail if `Review` were included in the clear).

### One thing worth knowing, not a defect

The orphan prune only runs on save. A long-lived session that closes a lot of panes keeps their entries in `pane_statuses` in memory until the next persistence write. Bounded by the write cadence and by the fact that `close_pane` panes are gone at the next save, so this is fine — just noting that "the map can't grow without bound" is true of the file, and true of memory only up to the next write.

Zeus-Deus:

I traced the persistence order through browser/OpenFlow pane stripping, live-pane collection, status filtering, layout serialization/restoration, and the existing activation semantics. The implementation correctly retains only `Review`, drops live-process states, and prunes orphan pane IDs.

GPT 5.6 Sol xhigh found no errors in this PR.

PR #240, opened 2026-08-03
Fix text selection contrast and consistency

Zeus-Deus:
## Summary

- add semantic text-selection foreground and background tokens
- apply the selection contract at the document root for consistent highlight inheritance
- document the DOM selection contract and renderer-owned exclusions

## Verification

- `npm run build`
- `CARGO_BUILD_JOBS=2 NODE_OPTIONS=--no-experimental-webstorage npm run verify`
- visually checked chat content, the composer, and settings in Cool and Warm palettes


Zeus-Deus:
The token design and the root-scoped contract look right (warm palette resolves correctly since `.theme-warm` doesn't touch `--accent-ember`, and the composer's transparent-textarea case is exactly why the foreground needs to be defined). One real regression though:

**Selected text in the editor pane goes dark-on-dark.** `EditorPane.tsx` enables `drawSelection()`, and CodeMirror's native-selection-hiding theme only resets `background-color` (to transparent, `!important`) on `.cm-line ::selection` — it never resets `color`. With highlight inheritance, the new `:root::selection { color: var(--selection-foreground) }` (dark ink, oklch 0.205) inherits into editor lines, so selected code renders dark ink over CM's own selection layer, which is `var(--accent)` (oklch 0.268 dark / 0.272 warm) — about 1.2:1 contrast, effectively invisible. Previously, selected editor text kept its syntax colors.

Suggested fix: opt the editor's foreground out in `src/lib/codemirror-theme.ts`, e.g. add `"& ::selection": { color: "currentColor" }` to the theme (in highlight pseudos, `currentColor` resolves to the originating text's color), or pick an explicit selection pair there next to the existing `.cm-selectionBackground` rule. The new DESIGN-SYSTEM.md section should then be accurate as written — right now the "CodeMirror keeps its own theme integration" exclusion only holds for the background channel, not the foreground.

Minor doc note: the root-scoped selector relies on the highlight-inheritance model; on engines using the legacy `::selection` matching model it silently no-ops back to the native highlight. Current WebKitGTK/WebView2 are fine, but worth a sentence in the doc.

Zeus-Deus:
Fixed in a4709742.

**Editor regression.** Confirmed and reproduced: with `drawSelection()` only resetting the native highlight's *background*, the root pair's foreground inherited into `.cm-line` and painted selected code as dark ink over CM's `--accent` selection layer. `src/lib/codemirror-theme.ts` now resets selected line text to `currentColor`, so selected code keeps its syntax colors exactly as before the token change.

Scoped to `.cm-line` rather than the whole editor on purpose: CM's own chrome (search panel inputs, tooltips) uses real native selection, so it should keep the app-wide pair. Verified both — drag-selected code keeps its syntax colors, the search field keeps ember + dark ink, prose selection unchanged.

**Docs.** The renderer-owned exclusions are now accurate per channel: xterm.js needs no opt-out at all (`.xterm` is `user-select: none`, selection is renderer-painted), CodeMirror needs the foreground opt-out because inheritance reaches it. Added the note that the root-scoped selector depends on the highlight-inheritance model and degrades to the native highlight on legacy-matching engines (both shipped WebViews implement inheritance).

Also added `src/lib/codemirror-theme.test.ts` covering the `.cm-line` opt-out and the root token pair, so a future theme edit that drops either channel fails the suite instead of shipping silently.

Verification: `cargo check`, `cargo test` (all green), `npm run check`, `npm run test` (3805 passed).

PR #243, opened 2026-08-03
Add provider-correct cross-provider skill support

Zeus-Deus:
## Summary

- Replace first-wins skill discovery with a cwd-scoped, provider-aware inventory spanning readable filesystem definitions, Codex skills/list, and OpenCode /skill.
- Carry exact collision-safe skill IDs through frontend selection and backend revalidation, with provider-native invocation where supported and provenance-aware portable fallbacks.
- Add per-provider compatibility projections, worktree-stable preferences, watcher invalidation, honest Settings controls, and deferred-worktree selection rebasing.
- Document provider contracts, queue snapshot semantics, rollout decisions, and the completed implementation.

## Review hardening

An independent high-reasoning code review drove fixes for project cache leakage, provider filtering drift, deferred-worktree ID changes, duplicate native names, catalog error handling and merges, provider-specific enabled state, whitespace preservation, watcher invalidation, worktree-stable disabled preferences, remote OpenCode content, envelope escaping, skill-name validation, collision suffixes, resolution latency, and exact Codex wire coverage.

## Verification

- npm run check
- npx vitest run --maxWorkers=2 — 250 files and 3,810 tests passed
- cargo check --manifest-path src-tauri/Cargo.toml --jobs 2
- cargo test --manifest-path src-tauri/Cargo.toml --jobs 2
- Fake Codex app-server wire test verifies exact structured skill items.
- Browser-based Settings smoke with the development Tauri mock.

All Cargo commands ran with CARGO_BUILD_JOBS=2.


Zeus-Deus:
Thorough piece of work — the exact-id + backend-revalidation model, the projection matrix, and the wire-level Codex test are all solid. Two things before this can land:

## Merge conflicts

The branch is conflicting with `main` in exactly two files, both docs:

- `docs/core/STATUS.md` — `main` (docs refresh, `430b9172`) rewrote the "Unreleased after v0.15.6" list (renumbered, PR-number annotations, new bullets) in the same region this PR inserts its bullet.
- `docs/INDEX.md` — same refresh touched the feature lines adjacent to the new cross-provider-skills entry.

A rebase onto `main` with those two hunks re-applied should be trivial; no code files conflict. (Note `main` has moved further since this analysis — several PRs merged today — so re-check after rebasing.)

## Code issues

1. **Not-installed providers render as errors** (`inventory.rs::harvest_codex_catalog` / `harvest_opencode_catalog` → `SkillInventory.errors` → the Settings warning box). A user with only Claude installed permanently sees "Some provider catalogs could not be refreshed … codex: codex_not_installed". Suggest treating a missing binary as an empty catalog and reserving the banner for real failures.
2. **`resolve()` hardcodes `include_plugins: true`** (`agent_chat_send_turn`). With the plugin toggle off, the popup caches under `(cwd, false)` while resolve looks up `(cwd, true)`, so every skill send after the TTL does a full forced discovery — including a fresh `codex app-server` spawn — and plugin-skill ids remain resolvable with the preference off. Passing the caller's actual preference through fixes both.
3. **Double catalog harvest on first skill send**: `resolve()`'s no-hint path runs `self.list(force=true)` and then rebuilds the inventory with per-provider harvests again. Reusing the forced list result would halve the spawns.
4. **`validate_skill_name` silently drops previously-visible skills** (`scanner.rs::build_skill`). Any existing skill with an uppercase/underscore/space name now vanishes from Settings with only a stderr line. Consider surfacing invalid entries (e.g. as unreadable/invalid rows) instead of dropping them.
5. **Dead `$name` branch**: `render_skill_invocations`'s Codex `NativeCommand` arm is unreachable — no projection ever assigns `NativeCommand` to a Codex target (always `CodexSkillItem`/`PromptPrefix`/`None`). The plan doc itself cautions against emitting `$name`; suggest removing the branch.
6. **Native Claude rendering swallows prefixes**: the frontend applies mode/attachment prefixes before send, and the backend then emits `/{skill} {entire prefixed text}`, so a plan-mode preamble or attachment block becomes the command's `$ARGUMENTS`. Worth deciding whether the native-command path should bypass or reorder `applyAllPrefixes`.
7. **Disabled project skills re-enable after upgrade**: persisted `disabledIds` carry old path-hash ids, but project skills now key preferences on the new canonical-repo `preference_id`, and no migration exists (the plan doc called for opportunistic migration).
8. Minor: `materializeWithPreset` doesn't thread `refreshSkillSelection`, so preset launches into a new worktree can hit `skill_not_found` on path-derived ids; and the ancestor project dirs scanned by `filesystem_inventory` aren't covered by the fs watcher (TTL-only freshness).

The `enumerate_scan_paths` sync-lockstep guards are untouched and still hold, and the new test coverage (wire capture, collision suffixes, worktree rebase, store race guards) looks good.

PR #244, opened 2026-08-03
Refine floating GUI chrome

Zeus-Deus:
## Summary

- consolidate workspace tabs and actions into a cleaner floating titlebar layout
- align sidebar, right-panel, resource-monitor, and native window controls
- raise titlebar control groups only when they physically overlap scrolled chat content
- preserve the transparent frameless state when the reading column is clear
- add regression coverage for responsive placement, control grouping, and content-under state

## Testing

- `npm run check`
- `NODE_OPTIONS=--no-experimental-webstorage npm run test` (251 files, 3,815 tests)
- visual checks at 1260x900 and 1800x900


Zeus-Deus:
The floating-overlay direction looks good and the token-based styling is clean, but I hit three functional issues while tracing the new code paths:

**1. Legacy (flag-off) chrome gets the overlay clearance without the overlay**

`sidebar-action-row.tsx` (`pt-11`, both expanded and collapsed) and `right-panel.tsx` (`mt-10`, gated only on `!isRemoteClient()`) apply the top-collision clearance unconditionally. With `enable_agent_chat` off — or on with an OpenFlow workspace — `TitleBar` renders the in-flow legacy `h-9` bar, so these surfaces already start below the chrome. Result: ~32px of dead padding above the sidebar search row, and a 40px blank band above the right-panel tabs when opened from the legacy TabBar. This contradicts the "every other case keeps the legacy chrome unchanged" invariant that `gui-chrome.md` still states. These paddings should be gated on the same predicate that puts the titlebar into overlay mode (note `right-panel.test.tsx` currently locks in the ungated `mt-10`).

**2. `useTitlebarChatOverlap` goes stale after the first tab/workspace switch**

The `useLayoutEffect` in `title-bar.tsx` depends only on `[enabled]` and snapshots `document.querySelectorAll('[data-slot="transcript-list"]')` once. `PaneContainer` renders only the active surface, so switching tabs or workspaces remounts `MessageList` with a fresh DOM node the hook never observes — and `guiChrome` stays `true` across those switches, so the effect never re-runs. After any navigation, `overlapsChat` can never become true again and the raised-island treatment silently stops firing (`contentUnder` alone isn't enough). The existing test only covers the initial-mount case (transcript appended before `render()`). Suggest keying the effect on the active workspace/tab, or having `MessageList` register its viewport element in the shared `titlebar-content-under` store so the titlebar always measures live nodes.

**3. Remote client: right-panel tabs sit under the pointer-eating overlay strip**

The overlay renders a full-width `pointer-events-auto absolute inset-0` drag div at `z-30` across the top 40px. On desktop the panel tab row clears it via `mt-10`, but on the remote client that clearance is skipped (`right-panel.tsx`) while the overlay div still renders — so the top 40px of the 45px tab header (i.e. the clickable centers of Files/Tasks/…) is covered and the tabs become effectively unclickable in GUI mode on the web client. The `mt-10` skip should only drop the *window-controls* allowance, not the overlay allowance.

Smaller notes:

- During a lazy draft, the floating band reads `rightPanelOpen` from the backend-active (previous) workspace; the draft never renders the panel, so the band can stop ~328px short of the right edge with the resource monitor floating mid-air.
- With `pt-10` dropped for a sole-root chat, the top 40px of the transcript is a click/wheel dead zone under the drag region (window-drag on desktop, nothing on web). If that's the intended tradeoff, fine — flagging so it's a deliberate call.
- `<Cpu />` in `ResourceMonitor` lost its explicit `h-3.5 w-3.5`; with `size="icon-sm"` the button base applies `size-4`, so the icon grows 14→16px in the legacy titlebar too.
- `gui-chrome.md` still claims flag-off chrome is "byte-identical", but the `FileDiff`→`PanelRight` swap in the legacy `tab-bar.tsx` and the flag-independent terminal pane restyle both change flag-off surfaces; worth softening that claim if these are intended.

Heads-up on concurrent work: #246 (legacy OpenFlow removal) edits the `workspace-main.tsx` region directly adjacent to the new `pt-10`/onboarding hunks and the same `gui-chrome.md` sections, so whichever lands second will need a careful rebase; #245 overlaps only trivially in `app-shell.tsx`.

Zeus-Deus:
## Review fixes — all three functional issues + every smaller note

Pushed as `ec0f00c5`. Each fix has regression coverage that I verified **fails against the previous behavior** before wiring it up.

### 1. Legacy chrome got the overlay clearance without the overlay ✅

Introduced `useTitlebarOverlay()` in `use-gui-chrome.ts` — `useGuiChrome() || useDraftGuiChrome()`, i.e. literally "`TitleBar` is rendering the floating overlay rather than the in-flow `h-9` bar". Both `sidebar-action-row.tsx` (expanded → `pt-3`, collapsed → `pt-2`) and `right-panel.tsx` now gate on it, so flag-off / OpenFlow keeps its original insets.

Confirmed visually in the dev mock with the GUI flag toggled off — before/after on the same seed data:

| before (`mt-10`/`pt-11` ungated) | after |
|---|---|
| sidebar search row at y≈96, panel tabs at y≈98 | y≈63 / y≈58, both flush under the legacy bar |

`right-panel.test.tsx` locked in the ungated `mt-10`, as you noted — it now mocks the predicate and asserts all three states (overlay+desktop → `mt-10`, overlay+remote → none, legacy → none).

### 2. `useTitlebarChatOverlap` went stale after the first switch ✅

Took the second suggestion: `MessageList` registers its scroll viewport in the shared `titlebar-content-under` store, which gained an element registry + monotonic version counter. `TitleBar` reads that version via `useSyncExternalStore` and adds it to the effect deps, so mount/unmount of any transcript re-runs the measurement against live nodes. Kept the `querySelectorAll` as a superset for any non-registering path.

The new test walks the actual failure: mount with transcript A → raised, unregister A (tab switch away) → dormant, register a **different** node B → raised again. On the old `[enabled]`-only deps it fails at step 2, proving the effect never re-ran.

### 3. Remote client: tabs under the pointer-eating strip ✅

Went one step further than dropping only the window-controls allowance: the drag layers themselves are now desktop-only. `data-tauri-drag-region` is inert in a browser, so on the web client that full-width `inset-0` layer was pure pointer sink — and the same applies to the calm gap between the islands. Dropping both fixes the tab row *and* frees the top 40px of the transcript on web, so the `mt-10` skip there stays correct (nothing left to clear).

### Smaller notes

- **Draft reads the previous workspace's `rightPanelOpen`** — fixed; `rightPanelOpen` is now `guiChrome && …`, matching how the draft branch already suppresses the other workspace-scoped controls. Test asserts the band stays at `right: 104px` with a draft active and the backend workspace's panel open.
- **`pt-10` dropped for a sole-root chat** — deliberate, and now written down as a constraint in `gui-chrome.md`: it's the standard frameless-app tradeoff for letting the transcript reclaim the top edge while keeping the window draggable. Note the web client no longer pays it at all after fix 3.
- **`<Cpu />` grew 14→16px** — good catch, restored the explicit `h-3.5 w-3.5` with a comment about `icon-sm` setting no `[&_svg]` override. Parameterized test covers both variants.
- **`gui-chrome.md` "byte-identical" claim** — softened. It now says flag-off *layout* is unchanged but calls out the two intentional flag-off changes by name (the legacy TabBar `FileDiff`→`PanelRight` swap, and the flag-independent terminal pane restyle).

### Verification

- `npm run check`
- `NODE_OPTIONS=--no-experimental-webstorage npm run test` — 251 files, 3,833 tests
- Visual: dev mock at 1280x633, GUI flag on and off, right panel open and closed
- No Rust touched

PR #247, opened 2026-08-05
Anchor new turns in the chat transcript (new-turn scroll contract)

Zeus-Deus:
## What

Replaces the chat transcript's one-shot send-to-bottom scroll with a durable **new-turn scroll contract**: a composer submission is treated as an explicit navigation intent. The optimistic prompt is anchored 16px below the transcript top, response space is reserved beneath it, and live follow persists until the reader deliberately navigates away.

## Why

The previous behavior incremented a scalar signal and called `scrollToEnd()` once while LegendList was still measuring the optimistic row and footer — one shot, uncertain landing, no durable follow mode, and a "Jump to latest" pill that could be left active by transient layout state.

## How

- **Three explicit states, one anchored identity** (`send-scroll-state.ts`): `following-end` (default), `anchoring-turn` (on send), `free-scrolling` (on gesture). `MessageList` owns them; nothing above it decides where the viewport goes.
- **Identity, not "last row".** `AgentChatPane.handleSubmit` issues a `sendAnchor` (`{ clientNonce, nonce }`) in the same batch as the optimistic append, reusing the bubble's existing `clientNonce`. The slot resolves by nonce (last match wins) because queued/control rows can land after the prompt. Text, image, one-click Continue, and queued "send now" turns all share the contract.
- **Anchor after measurement, never on a timer.** The resolved index feeds LegendList's `anchoredEndSpace` (16px offset); its `onReady` positions the measured row via instant `scrollToIndex`. While an anchor is mounted, the built-in `maintainScrollAtEnd` pin is disabled — two targets must not fight over one viewport. As the answer streams, the viewport advances by exactly the delta needed to reveal the tail (zero while the turn fits).
- **Opt-out is the reader's call.** Only real gestures (`wheel`/`touchmove`/`pointerdown`, subagent-jump) release follow — deliberately not `scroll`, which programmatic corrections emit. A generation counter guards every async continuation, so a gesture landing mid-flight wins. Failed sends clear the anchor on both rollback paths; thread switches reset to `following-end`.
- **Pill honesty.** Hidden synchronously on send, shown on a 150ms trailing debounce (raw `isAtEnd` reads false through mount/layout settling), and using it re-claims follow.

No dependency, backend, or virtualizer changes — the installed `@legendapp/list` 3.2.0 already ships `anchoredEndSpace`. The performance contract is preserved: stable slot keys, bounded DOM, 800px draw distance, zero transcript-row renders on composer keystrokes.

## Files

- `src/components/chat/MessageList.tsx` — scroll-intent owner: state machine, `anchoredEndSpace` wiring, gesture opt-out, pill visibility
- `src/components/chat/AgentChatPane.tsx` — issues/clears the send anchor beside the optimistic append and both rollback paths
- `src/components/chat/ChatTranscript.tsx` — forwards the anchor without weakening the memo boundary
- `src/components/chat/send-scroll-state.ts` (new) — pure state transitions + measured-turn geometry, testable without a virtualizer
- Tests: `send-scroll-state.test.ts` (new), `MessageList.test.tsx`, `MessageList.virtualization.test.tsx` (honestly windowed list double — the mounted range moves only when the component itself scrolls), `AgentChatPane.test.tsx`
- `docs/features/agent-chat.md` — transcript-scroller contract updated

## Verification

- `npm run check` — clean
- `npm run test` — 254 files / 3876 tests pass
- `cargo check` + `cargo test` — pass (one pre-existing load-flaky PTY flow-control timing test failed under full parallel build load and passes in isolation; no Rust files touched)
- Dev-mock browser: transcript hydrates pinned to bottom with no pill flash; send from deep history appends the bubble and reserves end space; gesture listeners attach to the scroller. Full scroll-follow interaction is not observable in the headless preview tab (it delivers no compositor frames, scroll events, or ResizeObserver callbacks — verified against the pre-change build, which behaves identically there), so viewport-position behavior is covered by the windowed-list integration tests and pure geometry units instead.


Zeus-Deus:
The contract itself is well designed — anchor-by-identity instead of last-row, measurement-driven positioning via `anchoredEndSpace.onReady`, and disabling `maintainScrollAtEnd` while an anchor is mounted are all the right calls, and the windowed-list test double is a real improvement. Verified the `@legendapp/list` 3.2.0 API surface and that the reserved end space shrinks to zero as the turn grows, so the geometry checks out. Two behavioral gaps need fixing before merge, plus two smaller lifecycle issues.

**1. Stale `isAtEndRef` defeats edge re-claim and pill scheduling (`MessageList.tsx:383-405`)**

The owner-swallow branch (`if (!atEnd && ownsScroll())`) returns without updating `isAtEndRef`, and `claimScroll` pins it `true`. During an anchored overflowing stream the real `isAtEnd` oscillates while every report is swallowed or deduped, so the ref stays `true`. If a wheel gesture lands during a not-at-end window (the frames between token layout and the double-rAF correction — common while streaming):

- reading history shows **no pill**: `scheduleJumpToLatest` only runs on an `isAtEnd` transition, and the value never transitions up there;
- scrolling back to the bottom emits `true`, which the `isAtEndRef.current === atEnd` dedup drops — the re-claim branch never runs, follow stays off, and the next token slides the transcript away.

That breaks the documented "scrolling back to the edge re-claims live follow". Suggested fix: keep the pill suppression in the swallow branch but still record `isAtEndRef.current = atEnd` (or resync the ref from `getState().isAtEnd` inside `cancelFollowForUserNavigation`), and add a test for gesture-during-stream → scroll-back-to-edge → follow resumes.

**2. `pointerdown` retires follow on every in-transcript click (`MessageList.tsx:420-439`)**

The comment scopes `pointerdown` to scrollbar drags, but the listener sits on the scrollable node, so plan Accept / approval buttons, tool-card expands, copy buttons, and text-selection starts all bubble to it and permanently cancel follow. Since the anchor stays mounted after a send, `maintainScrollAtEnd` is `false` and the effect is the sole driver — so approving a plan mid-run now freezes the viewport while the run streams on, recoverable only via the pill (which issue 1 can also delay). On main, the built-in pin followed through clicks. Suggest scoping pointerdown to actual scrollbar interaction (`e.target === viewport`) or dropping it, plus a test pinning "clicking an approval does not stop follow".

**3. Queued "send now" failure strands the anchor (`AgentChatPane.tsx:1961-1981`)**

`requestSendAnchor` fires before `agentChatSendQueuedTurnNow`, whose `.catch` only toasts. The bubble stays queued (no rollback), so a failed dispatch leaves the list anchored on a bubble that will never stream, with reserved blank space below — inconsistent with the rollback discipline on the two composer paths. Clear the anchor in the catch.

**4. Anchor never expires on turn completion**

It's only cleared on failure or thread switch, so (a) a `MessageList` remount on the same thread (subagent drill-in/out, workspace away/back) re-parks a long-completed prompt near the top instead of opening at the latest row — right behavior for an in-flight turn, stale intent for a settled one; and (b) the built-in `itemLayout`/`footerLayout` re-pins stay disabled for the rest of the thread's life after the first send, so a late-loading image that grows a row without a data change won't re-pin. Consider clearing the anchor when the turn settles.

Nits: `docs/core/STATUS.md` didn't get a cluster entry (convention per behavior PR), and the branch name will land in the default merge-commit message — a squash-merge with a clean message would avoid that.

Zeus-Deus:
All four addressed in edda8a25, each with tests verified to fail when the fix is reverted (suite now 3885 passing, tsc clean):

1. **Stale `isAtEndRef`** — reality is recorded on every listener event (including swallowed ones), and the gesture cancel resyncs from `getState().isAtEnd` and makes the pill decision there, since that's the moment ownership changes hands and the edge-triggered listener can't report an unchanged value. Gesture-mid-stream now shows the pill, and scroll-back-to-edge re-claims follow.

2. **`pointerdown` scope** — now only a pointerdown targeting the scroll container itself (scrollbar drag) releases follow; presses on rows (plan accept, approvals, expands, text selection) target descendants and keep following. Tests pin both directions.

3. **Queued "send now" failure** — the catch clears the anchor alongside the toast.

4. **Anchor expiry** — cleared on the falling edge of the combined live signal (`streaming || isSending`, seeded so mounting onto a settled thread isn't a completion). Expiry is lifecycle, not navigation: a reader already free-scrolling keeps their viewport and pill; when we still own scroll it hands back to `following-end` and the built-in end pin returns. Remount-after-settle opens at the latest row.

STATUS.md got the cluster entry and agent-chat.md documents the expiry rule and pointerdown scoping. Agreed on squash-merge to keep the branch name out of the merge commit.

Zeus-Deus:
Follow-up noted from the post-fix verification pass (non-blocking, merged as-is): the anchor-expiry watcher keys on the thread-level streaming signal, so a queued "send now" anchor can be wiped by the *previous* turn's settle during the ready→running dispatch gap — the queued prompt then streams with plain tail-follow instead of parking at the top (single tail-snap, no freeze or stranded space). A fix could key expiry to the anchored `clientNonce`'s own turn settling, or debounce the falling edge across the dispatch window. None of the current tests model the inter-turn ready gap.

PR #339, opened 2026-09-03
perf: cheaper chat hydration, activation deltas and an idle-friendly git sweep

Zeus-Deus:
## Summary

Follow-up to #338, measured on a real 274-workspace profile (271 chat panes, threads up to 10.8k events) in the dev build with the perf trace enabled. Stacked on #338 so the diff shows only these changes; retarget to `main` once #338 merges.

Where the remaining switch latency actually was:

| phase | before | after |
|---|---|---|
| backend activate (mutate+emit) | 0.2ms | 0.2ms |
| payload per switch | 581KB full snapshot | ~300B delta |
| cold replay, 10.8k-event thread (reducer) | 153ms | 38ms |
| idle git sweep tick, 203 checkouts | 830–1000ms | 400–620ms |

## Changes

- **Transcript index for replay** — the reducer walked the whole transcript per tool_result / tool_use / user turn (~2700 full scans on a 10.8k-event thread). A `WeakMap` index keyed by array identity, carried forward by two copy-on-write primitives. State shape unchanged; bench test guards the ratio.
- **Activation delta** — `activate_workspace` now emits a revision-stamped `ActiveWorkspace` delta instead of the full snapshot. The two side effects the snapshot used to carry (selection persistence, debounced layout persist) are scheduled explicitly. Other `activate_workspace` callers keep the full emit since they mutate more than the active id.
- **Change-gated git sweep** — inactive checkouts are fingerprinted from `.git` metadata (HEAD, target ref, index / packed-refs mtimes) and only re-read when it moved or a periodic full refresh is due. The active workspace is always read.
- **Startup** — one scrollback meta walk instead of one per restored session; `synchronous=NORMAL` for the WAL-mode database.
- **Renderer** — always-mounted components subscribe to the appState fields they read rather than the whole object; xterm no longer lands in the eager entry chunk through two hook imports.
- **Trace** — hydrate phases (read / liveness probe / replay) are sub-measured inside a workspace switch.

## What is still slow (not in this PR)

First visit to a long chat is dominated by the transcript read itself, not the reducer: `hydrate:read-full` is 1467ms for 10.8k rows (SQLite → JSON → IPC → parse). That needs a tail-first paged hydrate and is a separate change.

## Verification

- `npx tsc --noEmit` clean; `vitest run` 341 files / 5281 tests; `cargo test --lib` 3214 tests
- Dev build on the copied real profile: 21 palette-driven switches across the 7 largest threads + Ctrl+] cycles, traces exported via Settings → Copy performance diagnostics; selection persisted across restart; no duplicate emits on the control-socket and Ctrl+Tab paths.

Model: claude-fable-5-1 · Harness: Hermes


Zeus-Deus:
CI note: `ci.yml` only triggers on pull requests targeting `main`, so this stacked PR shows no checks while it targets `optimize-codemux-workspace-switching`. Retarget to `main` once #338 merges and the full matrix runs. Local equivalents at head `21b26cb2`: `tsc --noEmit` clean, `vitest run` 341 files / 5281 tests, `cargo test --lib` 3214 tests, all passing.

PR #403, opened 2026-09-20
fix(webview): recover the interface with Ctrl+Shift+R when it stops responding

### LINKED ISSUE #308: fix: provide a safe native reload path when the renderer fails
## Problem

Codemux can be left with a live backend and running agent/terminal processes but a blank, transparent window after the WebKit content process exits or stops rendering. In that state, resizing or refocusing the window does not recover the UI.

`Ctrl+R` cannot currently recover it. The frontend deliberately registers `blockReload`, `blockHardReload`, and `blockF5Reload` for `Ctrl+R`, `Ctrl+Shift+R`, and `F5`, and returns `true` to prevent the browser reload. More importantly, this interception lives in the renderer, so it cannot provide any recovery when that renderer is already unavailable.

This differs from desktop shells such as Electron apps that expose reload through the native/main process: the application shell can reload or recreate the web contents even when the frontend is unresponsive.

## Proposed direction

Add a native Tauri-side recovery reload for the main webview. The exact UX needs a deliberate decision; possibilities include:

- make `Ctrl+R` a native application accelerator that reloads the main webview;
- keep accidental reload protection and use a distinct recovery shortcut or native menu item;
- require confirmation during healthy operation, while allowing recovery when the renderer is known to be unavailable.

The recovery action must be owned by the native application process, not implemented as `window.location.reload()` or another frontend-only handler.

## Safety constraints

- Reload only the frontend webview; do not restart the Codemux process, backend, PTY daemon, provider sidecars, agents, or terminal commands.
- Verify that active terminal sessions reconnect with their output and input state intact.
- Do not forward the shortcut into the focused PTY.
- Avoid duplicate frontend subscriptions, event listeners, or panes after reconnection.
- Preserve persisted layout, active workspace/tab selection, drafts, and unsent chat input where supported.
- Ensure repeated reloads are idempotent and do not leak web-content processes.
- Keep the headless `codemux serve` surface unaffected.
- Handle Linux/WebKitGTK first, while defining sensible behavior on macOS and Windows.

## Acceptance criteria

1. A native action can recover the main UI without restarting the Codemux backend.
2. A long-running command started before the reload continues running and remains usable afterward.
3. The recovery still works when React cannot receive keyboard events.
4. Normal use does not make accidental reload/data loss easier without an explicit product decision.
5. Focused tests cover shortcut/action routing and verify that only the webview reload is requested.
6. The user-facing shortcut or menu action is documented in the keyboard shortcuts UI.

## Context

Observed on Linux/Wayland with `codemux-bin 0.19.2-1`: the main Codemux process, WebKit network process, provider sidecar, and running command remained alive, while no live WebKit web-content process remained and the window showed only the desktop through its transparent surface. There was no OOM event or renderer core dump. This issue is about providing a reliable recovery mechanism; the underlying renderer-exit cause may need separate instrumentation or investigation.

Zeus-Deus:
Closes #308.

## Problem

When the webview's renderer process dies or wedges, the Codemux window goes blank or stops responding while the backend, the PTY daemon, provider sidecars, agents and running commands all keep working. The page cannot run its own shortcuts at that point, so the only way out was restarting Codemux — which kills every running agent.

Renderer *crashes* already auto-reloaded (#355), but that path depends on WebKit reporting the process as terminated. A page that is alive-but-wedged, or one past the 3-per-minute auto-reload cap, stayed stuck. `Ctrl+R` and `F5` are deliberately swallowed (reverse-i-search in terminal panes, live keys in curses apps), so neither was available as a way back.

## Fix

`Ctrl+Shift+R` is now a recovery reload owned by the app process:

- **Linux** — grabbed on the GTK toplevel, before the key reaches the page or a focused pty, so it works with a dead or wedged renderer.
- **macOS/Windows** — runs through the keybind and the new `reload_main_window` command (a page that can dispatch a keybind is by definition alive).

Either way only the webview reloads: `WebviewWindow::reload()`, never a process restart. Panes re-attach to their live sessions on load.

It does not yank a working UI out from under anyone. While the renderer answers, the app process asks first — it emits a reload request, the page acks it and shows a confirmation dialog, and a **second press reloads immediately** (the escape hatch for a page that runs JS but cannot paint). A page that stays silent for a second is wedged, and the watchdog reloads it without asking. `Ctrl+R` and `F5` keep being swallowed on a live page and still reload only a known-dead renderer.

The action is listed in Settings → Shortcuts and in the command palette ("Reload interface"). In the web remote client the chord stays the browser's own hard reload, and neither new command is on the remote allowlist.

## Verification

Unit tests cover shortcut classification, the auto-reload cap, the watchdog/ack ordering, and that the frontend routes the action to exactly one command (`reload_main_window`) and nothing else. Live run against `tauri:dev` on Wayland:

| Scenario | Result |
| --- | --- |
| Healthy page, one press | app process asks, dialog opens, no reload |
| Second press | reloads via `reload_main_window` |
| Cancel, then press again | asks again instead of reloading |
| Ack suppressed (wedged page) | watchdog reloads after ~1s |
| Renderer killed | auto-reload, then "auto-reload paused" at the cap |
| Dead renderer + Ctrl+R | reloads, fresh renderer spawns |
| Healthy Ctrl+R / F5 | nothing happens |

A `while true` loop started in a terminal pane before the reloads kept ticking without a gap across all of them (tick 14 → 156), with the same shell PID and the same app PID, and never more than one web-content process — no restarts, no leaks.

## Screenshots

New confirmation dialog:

Reload confirmation dialog

Settings → Shortcuts, before and after:

| Before | After |
| --- | --- |
| Before | After |

Model: Claude Opus 5 · Harness: Claude Code


Zeus-Deus:
Superseded by #406, which covers this PR's recovery goals natively on Linux and Windows, and now also includes the "Reload interface" command-palette entry from this PR. Closing.

